PR #70, opened 2010-12-08
Twitter's Improvements to Mustache.js

bcherry:
Hello!

As you know, we use Mustache.js on the new Twitter.com.  We've had to make a few changes to support our use, which I've just finished getting officially integrated into our fork on GitHub, with tests in the Mustache.js suite.  We'd love to be able to port these changes upstream to the wider community.

The main changes we've made are:

1) Added i18n support, through {{_i}}{{/i}} tags.  The README in our fork has more info on how this works.

2) Fixed a bug we uncovered recently, where the content of section tags would get rendered twice, allowing Mustache injection.  We fixed this silently since it was a pretty embarrassing bug in the site, the fix just hit our production site today.  I'd be happy to provide more details on this, but we've also added a regression test to the suite in our fork.

3) I've added Rhino as a test runner, in addition to running with jsc.  We actually found some subtle regex-related bugs that caused failures in Mustache in Rhino and v8 only, so more coverage here seemed warranted.  In the longer run, it would be nice to be able to also run the suite on a web page with QUnit or something, so you can try it in all browsers you'd like.

For now, we've changed the version to "0.3.1-twitter" to make it clear in our codebase, but we'd love to integrate with the main branch.

Thanks!

### DIFF
--- a/mustache.js
+++ b/mustache.js
@@ -187,35 +187,41 @@ var Mustache = function() {
         var renderedBefore = before ? that.render_tags(before, context, partials, true) : "",
 
         // after may contain both sections and tags, so use full rendering function
-            renderedAfter = after ? that.render(after, context, partials, true) : "";
+            renderedAfter = after ? that.render(after, context, partials, true) : "",
 
-        var value = that.find(name, context);
-        if(type == "^") { // inverted section
-          if(!value || that.is_array(value) && value.length === 0) {
+        // will be computed below
+            renderedContent,
+
+            value = that.find(name, context);
+
+        if (type === "^") { // inverted section
+          if (!value || that.is_array(value) && value.length === 0) {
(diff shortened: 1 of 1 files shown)


adjenks:
@bcherry 
Is the i18n stuff ready? I was hoping to use it. I was sad to see it got removed and never put back.

bcherry:
@adjenks sorry, unfortunately I no longer work at Twitter nor maintain any of their open-source projects. It does look like the Mustache.js fork was removed and I believe is no longer in use there. I haven't used it, but this NPM module appears to offer the same functionality in a somewhat more modern package: https://www.npmjs.com/package/mustache-i18n

adjenks:
@bcherry Thank you for the suggestion. Best of luck in whatever you're doing now.

PR #99, opened 2011-05-23
Mustache Spec compliance

pvande:
Since Mustache implementations should strive to be template compatible, this change adds basic compliance testing against the formalized Mustache Spec.  This is currently set up against v1.1.2 of the spec, without support for the optional lambda module.


phillipj:
Closing this as #301 was merged which fixed testing against the spec

PR #111, opened 2011-08-07
Content truncation which ends with "..."

akoenig:
In my current project it was necessary for me to truncate the strings within a template to a specific length and add the omnipresent "...".

I thought it would be cool if there is a notation for mustache tags to handle this problem. So I've implemented a small modification of the "render_tags" method. With this it is possible to shorten the string by defining the length within the tag:

{{thetag:20}} -> Shorten the "thetag" content to 20 chars and end up with " ...".

So maybe this is useful for somebody.

Regards,
André

### DIFF
--- a/mustache.js
+++ b/mustache.js
@@ -191,7 +191,7 @@ var Mustache = function() {
             break;
         }
 
-        if (truncate.shouldI && found) {
+        if (truncate.shouldI && found && found.length > truncate.after) {
             found = found.substring(0, truncate.after) + ' ...';
         }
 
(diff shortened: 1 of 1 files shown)


nikolas:
I'm also trying to make a truncation filter using Mustache.js 2.1.3.

How do I pass in a length argument to the filter in the Mustache template? Here's
what I have so far:

```js
    'ellipses': function() {
        return function(text, render, args) {
            console.log(args);  // undefined :P
            var length = 3;
            var str = render(text);
            if (str.length > length) {
                str = str.substring(0, length) + '&hellip;';
            }
            return str;
        }
    },
```

Issue #241, opened 2012-08-05
Dynamic Partials


ghost:
**Update: please see pull request for new solution: https://github.com/janl/mustache.js/pull/242.**

Hi,

I might be missing something obvious but can't think of a way to dynamically render a partial based on a value. For example, given the view below:

``` javascript
{
  items: [
    { type: 'image', title: 'Some title' },
    { type: 'text', content: 'Some text' }
  ]
}
```

I want to dynamically render a partial based on the 'type' key:

``` mustache
{{#items}}
   {{> {{type}} }}
{{/items}}
```

This obviously won't work but I thing there needs to be some way of doing this. Not sure what the syntax would be, something like the following(?):

``` mustache
{{#items}}
   {{@type}}
{{/items}}
```

or even...

``` mustache
{{>items:type}}
```

Currently, the only way to achieve this that I can see is to do the following:

``` javascript
{
  items: [
    { type: 'image', title: 'Some title', is_image: true },
    { type: 'text', content: 'Some text', is_text: true }
  ]
}
```

``` mustache
{{#items}}
  {{#is_text}}
    {{>text}}
  {{/is_text}}
  {{#is_image}}
    {{>image}}
  {{/is_image?}}
{{/items}}
```

Someone else has come up against the same problem here: http://stackoverflow.com/questions/2932679/dynamically-render-partial-templates-using-mustache

Kind regards,

Jamie

PR #242, opened 2012-08-06
Dynamic partials and partial collections

ghost:
Hi,

This is a small patch that brings great flexibility when dealing with collections that are going to be rendered using partials. Please see the new documentation below:
- https://github.com/thelucid/mustache.js#dynamic-partials
- https://github.com/thelucid/mustache.js#partial-collections

With this patch, the following:

View:

``` javascript
{
  items: [
    { type: 'image', url: 'Some URL', is_image: true },
    { type: 'text', content: 'Some text', is_text: true }
  ]
}
```

Template:

``` mustache
base.mustache
{{#items}}
  {{#is_text}}
    {{>text}}
  {{/is_text}}
  {{#is_image}}
    {{>image}}
  {{/is_image?}}
{{/items}}

text.mustache
<p>{{content}}</p>

image.mustache
<p><img src="{{url}}"/></p>
```

...can be replaced with:

View:

``` javascript
{
  items: [
    { partial: 'text', content: 'Some text' },
    { partial: 'image', url: 'Some URL' }
  ]
}
```

Template:

``` mustache
base.mustache
{{@items}}

text.mustache
<p>{{content}}</p>

image.mustache
<p><img src="{{url}}"/></p>
```

...or if more fine grain control is needed:

Template:

``` mustache
base.mustache
{{#items}}
<p>{{>.}}</p>
{{^items}}

text.mustache
{{content}}

image.mustache
<img src="{{url}}"/>
```

Patch includes tests and documentation, addressing a recurring problem when using Mustache templates e.g. http://stackoverflow.com/questions/2932679/dynamically-render-partial-templates-using-mustache and https://github.com/janl/mustache.js/issues/241.

I really hope the pull request is accepted as I am already utilising this in a couple of projects and it's a joy to use. I will look at implementing in the Ruby version of Mustache also if accepted.

Kind regards,

Jamie


vgracia:
This seems like a great feature.  Was this functionality ever added?

independentgeorge:
I know I'm late to the party, but in case it helps someone - here's a simple function enabling dynamic partials on syntax level without polluting the view.

Parentheses in partial's name represent the dynamic property whose value determines the template to use.

```javascript
// For every dynamic partial, a function which renders the subtemplate is added to the view's 
// prototype, and their occurrences in the template are replaced with `{{{partial-(name)}}}`, 
// now accessible through the prototype.

function render( template, view, partials = {} ){
   const prototype = {}
   template = template.replace(/\{\{>(.*?\(.+?\).*?)\}\}/g, function(_, name, property){
      if( !prototype[name] ){
         prototype[name] = function(){
            const computed = name.replace(/\((.*?)\)/, (_, property) => this[property])
            if( !partials[computed] )
               throw new Error(`Dynamic partial "${computed}" does not exist.`)
            return render(partials[computed], this, partials)
         }
      }
      return `{{{${name}}}}`
   })
   view = Object.assign(Object.create(prototype), view)
   return Mustache.render(template, view, partials)
}
```

```javascript
const template = "{{#items}}{{>(type)}}{{/items}}"

const partials = {
   image: '<img title="{{title}}">',
   text:  '<p>{{content}}</p>'
}

const view = {
   items: [
      { "type": "image", title: "Some title" },
      { "type": "text", content: "Some text"}
   ]
}

render(template, view, partials)  // <img title="Some title"><p>Some text</p>
```

Just make sure you're not using parentheses in property names of the view.

stuartjnelson:
@monkape thanks for sharing your solution. I can't get it to work. Do you happen to have any repos/live code examples using your solution you'd be happy sharing?

monkape:
@stuartjnelson Looks like the example above still works fine [https://jsbin.com/zejewasebe/edit](https://jsbin.com/zejewasebe/edit?js,output)

stuartjnelson:
Thanks @monkape for creating the demo. I need to spend some more time getting it to work with my stack. 

Express app which renders the templates based on JSON data and a `.mst` page. This page is made up of `.mst` partials. I was hoping I could get this magically working but it's proved unsuccessful so far.

lucas42:
To anyone still interested in the original suggested change here: I've done a bit of digging and it seems the change was proposed on the mustache spec.  After several iterations, it was merged into the spec as "Dynamic Names" a full 10 years later.
The syntax is significantly different to this original proposal, so I've created a fresh PR for the new syntax: #828

PR #261, opened 2012-10-01
Added piped helpers

techhead:
Implementation of piped helpers based on the discussion found here: https://github.com/janl/mustache.js/pull/139

See the new tests.

``` javascript
({
    numbers: [1,2,3,4],
    sum: function() {
        var total = 0;
        for (var i=0,len=this.length; i<len; i++) {
            total += this[i];
        }
        return total;
    }
})
```

``` html
{{numbers|sum}}
```

``` javascript
({
    timestamp: 1,
    iso8601Date: function() {
        return new Date(this).toISOString();
    }
})
```

``` html
{{timestamp | iso8601Date}}
```

Helpers are simply lambdas.

```
{{a | b}} is identical to {{#a}}{{b}}{{/a}} when 'a' is a single value
```

It is not identical when 'a' is a list.  The list will become the context of the lambda to allow for 'reduce' type functions.


raphaelkieling:
It's working? i did need and create this:

```js

// Split and trim pipes
function splitPipe(value){
    if(nullOrUndefined(value)) return []

    const pipe = "|";
    let trim = (string) => string.trim();

    if(value.includes(pipe)) return value.split(pipe).map(trim);

    return [value];
}

// Call the functions, `pipelineFunctions` is a object with funcions, because i create default funcions.
function pipe (value, pipes){
    if(nullOrUndefined(value)) return value;
    if(nullOrUndefined(pipes)) return value;

    return pipes.reduce((ant,functionName)=>{
        if(pipelineFunctions.hasOwnProperty(functionName))  return pipelineFunctions[functionName](ant);
        return ant;
    },value)
}

Mustache.Writer.prototype.escapedValue = function escapedValue (token, context) {
    let [,tokenValue]    = token
    let [name, ...pipes] = splitPipe(tokenValue)
    let contextValue     = context.lookup(name);

    let finalValue       = pipe(contextValue, pipes)

    if (finalValue != null)
      return Mustache.escape(finalValue);
};
```

PR #304, opened 2013-03-17
Mustache specs - lambda compliance

danielfagerstrom:
Fixed most of the failing lambda tests from the Mustache spec from my Mustache spec [pull request](http://github.com/janl/mustache.js/pull/301).

The [spec](https://github.com/mustache/spec/blob/master/specs/~lambdas.yml) says that for a function that is returned in a section or an interpolation:
1. When used as the data value for an Interpolation tag, the lambda MUST be
   treatable as an arity 0 function, and invoked as such.  The returned value
   MUST be rendered against the default delimiters, then interpolated in place
   of the lambda.
2. When used as the data value for a Section tag, the lambda MUST be treatable
   as an arity 1 function, and invoked as such (passing a String containing the
   unprocessed section contents).  The returned value MUST be rendered against
   the current delimiters, then interpolated in place of the section.

In current mustache.js the return value from a function in a interpolation or section is not interpolated so I implemented that.

For the section functions this breaks one of the current tests so I updated that one.

The test for "Section - Alternate Delimiters - Lambdas used for sections should parse with the current delimiters." still fails, I couldn't find any reasonably simple way to fix that.

### DIFF
--- a/README.md
+++ b/README.md
@@ -348,7 +348,7 @@ These may be built using [Rake](http://rake.rubyforge.org/) and one of the follo
     $ rake jquery
     $ rake mootools
     $ rake dojo
-    $ rake yui
+    $ rake yui3
     $ rake qooxdoo
 
 ## Testing


andrewnicols:
I've encountered this recently. What's the current state of play if I were to take on the remaining changes?

MiniHeyd:
Is there any update on this? I am still seeing this issue with lambda's where `{{ >second }}` is printing Instead of the partial it is supposed to render.

PR #315, opened 2013-05-23
fix for environments with changes in Object.prototype

nagaozen:
Although it's not recommended to, but it's allowed to augment Object.prototype with more properties and methods. Using hasOwnProperty enables Mustache to work with no changes in those environments.

### DIFF
--- a/mustache.js
+++ b/mustache.js
@@ -205,7 +205,9 @@
           self._loadPartial = partials;
         } else {
           for (var name in partials) {
-            self.compilePartial(name, partials[name]);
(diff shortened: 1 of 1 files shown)


dasilvacontin:
Hi! After a refactor, we now compile partials on demand, so there is no `for in` loop anymore. I would have merged, but there is no loop anymore, sorry. :/

nagaozen:
If it works out-of-the-box, not merging is not a problem xD

dasilvacontin:
Yeah, but you took the effort to make the PR, so it's a shame. :)

PR #347, opened 2013-12-05
loaded as worker detection and support

seacgroup:
starting at line 567, detect if running as worker, if so add message listener


nikelborm:
This PR can most likely be closed. PR's author hasn't been online for a long time.

PR #349, opened 2013-12-11
mustache-cli in coffeescript

robblovell:
I created a mustache-cli for mustache.  I haven't looked at it since August, and it may need a bit of love to make it better.  But, for what it's worth, here it is.  Written in coffeescript, so hopefully, that's not a problem.

Robb

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -7,4 +7,5 @@ qooxdoo.mustache.js
 dojox
 yui3
 requirejs.mustache.js
+.idea/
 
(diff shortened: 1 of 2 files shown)


dasilvacontin:
Wow. I'm sorry for this @robblovell, but we recently decided to create a CLI and we overlooked this PR.

Still, the spirit of mustache.js is that it is a very simple approachable Javascript implementation of mustache. It would make sense to also code the CLI in Javascript, not in Coffee.

PR #358, opened 2014-02-18
Add a filter for emitting JS data

balihoo-anewman:
In our application, we sometimes need to embed Javascript variables:

```
<p>Hello, {{name}}.</p>
<script>
  var x = {{{x}}};
</script>
```

However, in order to do this, we've had to call `JSON.stringify` from the calling code before rendering the template:

```
mustache.render('...', {
  name: 'thomas',
  x: JSON.stringify([1, 2, 'asdf']),
});
```

It's easy to forget which variables must be stringified, and which shouldn't be.  This is because mustache is designed to emit HTML.  However, JS is such an integral part of HTML, it deserves its own emitting mode.

Additionally, JSON.stringify actually can't even be relied upon to emit correct JS (http://timelessrepo.com/json-isnt-a-javascript-subset).  The string `JSON.stringify('\u2028')`, for instance, is an invalid JS string, because it contains a line separator.

This commit implements a new filter which will emit JS.  The above template would be written:

```
<p>Hello, {{name}}.</p>
<script>
  var x = {{js&x}};
</script>
```

The `js&` tag was chosen arbitrarily.  Any other tag would work just as well.

### DIFF
--- a/mustache.js
+++ b/mustache.js
@@ -73,6 +73,10 @@
           s += "\\'";
         } else if (c == '\\') {
           s += '\\\\';
+        } else if (c == '<') {
+          s += '\\x3c';
+        } else if (c == '/') {
+          s += '\\x2f';
         } else if (n < 32 || n > 122) {
           var code = n.toString(16);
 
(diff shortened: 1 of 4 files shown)


brandonros:
What is the proper way to do this now?...

Mygod:
An alternative way to do this: `JSON.stringify(JSON.stringify([1, 2, 'asdf']))`.

PR #359, opened 2014-02-20
Custom tags

zverev:
Hello,

I've tried to make an internationalization for our project using mustache, but didn't find a rather simple way to do it. So i've thought about using custom tag for this task. For example `{{i some text}}`.
Does this idea corresponds to the philosophy of mustache.js?

Best regards,
Alexander

### DIFF
--- a/test/custom-tag-test.js
+++ b/test/custom-tag-test.js
@@ -1,0 +1,86 @@
+require('./helper');
+
+describe('Add custom tag function', function () {
+  describe('for an non-alphabetic tag string', function () {
+    it('throws an error', function () {
+      assert.throws(function() {
+        Mustache.addCustomTag('tag?', function () {});
+      }, 'invalid tag name');
+      assert.throws(function() {
+        Mustache.addCustomTag('0tag', function () {});
+      }, 'invalid tag name');
+    })
+  });
+
+  describe('for an illegal function argument', function () {
+    it('throws an error', function () {
+      assert.throws(function() {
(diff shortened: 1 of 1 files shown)


zverev:
@dasilvacontin 
According to this pull-request:
```javascript
var translationsHash = {
    firstName: "First name",
    lastName: "Last name"
};

Mustache.addCustomTag('i', function(contents, value) {
    return translationsHash[contents];
});
```
Template may look like this:
```html
<div class="user">
    <div class="user-firstName">{{i lastName}}: {{firstName}}</div>
    <div class="user-lastName">{{i firstName}}: {{lastName}}</div>
</div>
```
If we use this approach, we don't need to pass strings at all.

@phillipj 
Actually Handlebars has this feature (helpers), so we switched to it. Now I'm not sure, that mustache.js really need custom tags.

dasilvacontin:
@zverev, thanks for the template example.

Couldn't you just pass a different data/model object or a modified one to `Mustache.render`?

zverev:
@dasilvacontin 
Yes, I certainly could. But this method allows to avoid it. Also in my case I have some global translations, that are used by multiple components (e.g. button titles such as 'ok', 'cancel'). Without 'i' tag I would have to pass them manually.
Sorry for the late reply.

dasilvacontin:
No worries. I see what you mean.

We certainly need some kind of Plugin API. Having a Plugin API would make solving this issue something trivial. I think we should discuss how we approach adding a Plugin API in a separate issue.

PR #388, opened 2014-07-01
Backtick is added to entityMap

mtimur:
Backtick is not being HTML encoded currently. I added it to entityMap.

Backticks may be used to perform XSS on some certain situations for IE8.0 clients, for e.g. http://html5sec.org/#102


dasilvacontin:
Thanks for the report, @mtimur. I'll try to research the topic, read the paper, and follow up this PR.

Issue #397, opened 2014-11-08
Looking up in parent context on foo: null is contrary to RFC


Roman600:
Mustache RFC:
>Variables
The most basic tag type is the variable. A {{name}} tag in a basic template will try to find the name key in the current context. If there is <b>no name key</b>, the parent contexts will be checked recursively. If the top context is reached and the name key is still not found, nothing will be rendered.

But <b>I do have</b> a property in children object and it's value is <b>null</b> (which is false in some way). And I expect nothing to be rendered, but instead mustache.js looks up for a value in parent context and finds it, and shows it. But it's parent's value, not children's. And children <b>has it's own value</b>, and it's null. So according to RFC looking up doesn't need to be done.

The same thing happens not with only variables {{key}}, but with sections {{#key}}{{/key}} too.

phillipj:
Fixed in a020430

dasilvacontin:
Oh, that commit is from when I believed writing 'fixes' once before a bunch of issue numbers was enough. :smile:

PR #407, opened 2014-12-18
fixing null and undefined lookup

dasilvacontin:
Issue #397.


dasilvacontin:
Regarding backwards compatibility, what are your thoughts, @phillipj, @janl?

IMO, this could be considered a bugfix/patch, that might even help someone notice tests that should be failing.

phillipj:
Although I completely agree that this should be considered a bugfix, its a change which might cause  unexpected behavior in many projects. Being very explicit about this change for those developers upgrading to v2.0 would hopefully save many hours of debugging.

I really appreciate good changelogs / upgrade guides when libraries do breaking changes. Could that be an idea for the v2.0 release?

janl:
Even if an issue fixes a bug, if it breaks BC, it breaks BC and should be treated as such (carefully, major version bump).

Good upgrade guides are important too.

I’m a fan of small-scope major version releases, so it is easy to upgrade (like hapi generally does).

Major version numbers are cheap.

dasilvacontin:
True, sorry for missing that. Thanks guys! It is more likely to change behaviour in unexpected ways in lots of projects, than helping to discover old bugs.

I label this PR as v2.

dasilvacontin:
Heh, I love edge cases. Not only the `if` was unreadable, but it was failing with this edge case:

```javascript
> view = { shares: { fb: 5 }, kipos: { shares: { fb: undefined } } }
> Mustache.render("{{#kipos}}{{shares.fb}}{{/kipos}}", view)
'5'
```
This is due to `context.view.hasOwnProperty(name)` failing when using dot notation.

dasilvacontin:
I've refactored the if using a new variable, `lookupHit`. Can you guys please take a look and give feedback? :)

phillipj:
Once again, awesome tests! LGTM

dasilvacontin:
Thanks! I'll mark this as `TO-MERGE` and `v2.0`. Coming soon.

Issue #411, opened 2014-12-24
refactoring some large functions


dasilvacontin:
As I mentioned in #406, refactoring some functions would make the codebase easier to maintain, understand and extend. I'm specially looking at parseTemplate (~125 lines) and Writer.prototype.renderTokens (~100 lines).

PR #414, opened 2015-01-07
Refactor Writer.renderTokens() for better readability.

phillipj:
Hi,

PR for #411. It splits the old Writer.renderTokens() into several small functions with a single responsible.
Also removed a lengthy case-switch which quickly gets error prone.

Any thoughts?

### DIFF
--- a/mustache.js
+++ b/mustache.js
@@ -444,84 +444,96 @@
   Writer.prototype.renderTokens = function (tokens, context, partials, originalTemplate) {
     var buffer = '';
 
+    var token, tokenRenderer;
+    for (var i = 0, numTokens = tokens.length; i < numTokens; ++i) {
+      token = tokens[i];
+      tokenRenderer = this._getTokenRenderer(token[0]);
+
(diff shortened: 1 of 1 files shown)


phillipj:
Seems the alternative case-switch added a measurable performance hit :( I'll see what I can do with it.

dasilvacontin:
Keep us updated!, and thanks for the work! :)

dasilvacontin:
Any update @phillipj?

phillipj:
Sorry @dasilvacontin, forgot to push some local changes.. By doing a more similar if-else style comparable to the old switch-style, the performance penalty disappeared.

Would you considered it different enough to see any value in this refactoring?

dasilvacontin:
No worries, I was just curious! :) It definitely looks cleaner. I will take a better look tonight!

dasilvacontin:
:+1:

dasilvacontin:
Oups, commits were not squashed.  I have to be extra-careful about that next time.

PR #418, opened 2015-01-08
Drop `#` from `tag` comment

gjtorikian:
Jekyll, the static site generator, is planning to move to the newest version of Liquid, Liquid 3, running in `strict` mode.

It seems that Liquid really hates the fact that there's a non-alphanumeric tag in mustache.js. It interprets the `{{#tag}}` comment badly. This PR removes the `#` from the comment.

### DIFF
--- a/mustache.js
+++ b/mustache.js
@@ -94,7 +94,7 @@
     var nonSpace = false;  // Is there a non-space char on the current line?
 
     // Strips all whitespace tokens array for the current line
-    // if there was a {{#tag}} on it and otherwise only space.
+    // if there was a {{tag}} on it and otherwise only space.
     function stripSpace() {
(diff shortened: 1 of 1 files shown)


bobthecow:
Why is liquid parsing mustache.js?

gjtorikian:
Mustache is being pulled in because I'm using https://github.com/slashdotdash/jekyll-lunr-js-search, which requires it. Though as I'm typing it out, maybe the better fix would be to move search off of the second templating language...

dasilvacontin:
IMO, it should just be told not to parse the tags. Not being able to write comments about sections in the source code just because X tool messes up or you are not configuring it correctly is not the way to go. I'll reopen otherwise.

PR #421, opened 2015-02-02
Run jshint before tests

phillipj:
Hi,
`jshint` should also be able to break the build, not just broken tests. Atm `jshint` run as a part of the `Rakefile`, that's a bit too late. It forces the maintainer doing a new mustache.js release to clean up codestyle violations. IMO it should be the PR developer's responsibility to clean up the code.

This PR includes a good example; clumpsy me left an unused variable in my previous refactoring PR #414. That might have broken the release script and someone else would have to fix my mistake. I would ofcourse have fixed it immediately if the travis run would have failed.

### DIFF
--- a/.jshintrc
+++ b/.jshintrc
@@ -1,5 +1,6 @@
 {
   "eqnull": true,
-  "evil": true
+  "evil": true,
+  "unused": true
 }
 
--- a/.travis.yml
+++ b/.travis.yml
@@ -4,6 +4,8 @@ node_js:
   - 0.8
   - '0.10'
   - 0.11
+before_install:
+  - "test $TRAVIS_NODE_VERSION = '0.6' && npm install -g npm@1.3.26 || npm install -g npm"
 matrix:
   allow_failures:
     - node_js: 0.11
--- a/README.md
+++ b/README.md
@@ -490,6 +490,14 @@ Then, you can run the test with:
 
     $ TEST=mytest npm run test-render
 
+### Troubleshooting
+
+#### npm install fails
+
+Ensure to have a recent version of npm installed. While developing this project requires npm with support for `^` version ranges.
+
+    $ npm install -g npm
+
 ## Thanks
 
 mustache.js wouldn't kick ass if it weren't for these fine souls:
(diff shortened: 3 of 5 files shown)


dasilvacontin:
It's failing on older npm versions, but a very big :+1: to adding `jshint` to the test process.

I think the same happened with mocha, and I pinned the version... I should really look into a real solution.

phillipj:
Huh.. Which old version of npm fails? Maybe its as simple as carrot vs
tilde for the version ranges.

dasilvacontin:
I'm not sure which version changed the behaviour, or with which version it started failing. Only that it's working with 1.4.x, whilst with < 1.4 it's not working. Maybe we should enforce an npm version? I should really read more about this.

phillipj:
Yupp, its a carrot issue :( The short version is that the old tilde is more strict, only allowing patch version upgrades. Carrot allows minor and patch upgrades. The tricky part is that the *old* versions of npm, e.g. the ones bundled with node v0.6 and v0.8, doesnt like carrots at all.

I'll do some digging and see if I can find a way around getting jshint installed, as it depends on other modules which again uses carrot for their deps..

P.S. mocha seems to work fine with older node versions when declared with tilde.

dasilvacontin:
Nice to know! If you find any interesting links, please share. :)

phillipj:
Found a few references and one common solution: upgrading npm before running the tests by adding this to `.travis.yml`

```yaml
before_install:
  - npm install -g npm
```

- https://github.com/npm/npm/issues/4849
- http://tech.vg.no/2014/05/30/npm-travis-node-and-the-caret-operator/

Newer npm versions are supposed to work with older versions of node, therefore we can upgrade npm and still run tests on node v0.8 and early v0.10's which has npm < v1.4.0 bundled. After lots of trial and error I even got it running with node v0.6 by using a special combination of npm v1.3.26 and jshint 2.4.4.

As there's suddently a required minimal version of npm installed when running tests, I've added a troubleshooting section in the readme and added `engines.npm` field in `package.json`: https://docs.npmjs.com/files/package.json#engines.

dasilvacontin:
@phillipj, you are da real MVP! Squash commits and I will gladly merge this. :+1: 

Thanks for providing links about your findings. :)

phillipj:
Thanks, just happy to help :)

dasilvacontin:
:tada:

Issue #415, opened 2015-01-07
number rendered as [object Object]


coolaj86:
I'm using Mustache `"mustache": "~0.8.2"` via bower in Chrome `Version 39.0.2171.95 (64-bit)`

I have this in my template

```html
  {{# data.author.twitter_id }}
    <meta name="twitter:site:id" content="{{ . }}">
  {{/ data.author.twitter_id }}
```

and this is my data

```javascript
{ data: { author: { twitter_id: 15191378 } } }
```

So this is literally run as

```javascript
output = Mustache.render(
  '  {{# data.author.twitter_id }}\n   <meta name="twitter:site:id" content="{{ . }}">\n   {{/ data.author.twitter_id }}'
, { data: { author: { twitter_id: 15191378 } } }
);
```

And literally gets this output:

```html
<meta name="twitter:site:id" content="[object Object]">
```

What!?!?!?!?!

Any ideas?

Now, I can converts numbers to strings before passing them to Mustache, but that may actually turn out to have a lot of hard-to-catch cases in my application. I'd imagine this is a fairly simple type check fix somewhere in the bowels of the Mustache code?

bobthecow:
It's something to do with the implicit iterator tag (`{{ . }}`). This template does what you'd expect:

```html+jinja
{{# data.author }}
    <meta name="twitter:site:id" content="{{ twitter_id }}">
{{/ data.author }}
```

I don't know where to fix it, but hopefully that helps someone narrow it down :)

dasilvacontin:
I see, interesting bug. I wonder if fixing this will break other test cases.

rndme:
expanding renderTokens with number handling seems to fix it: `  } else if (typeof value === 'object' || typeof value === 'string'|| typeof value === 'number') {`

phillipj:
Spot on @rndme :+1: I just opened a pull request with the fix you proposed

PR #428, opened 2015-02-17
Bugfix for number rendered as [object Object]

phillipj:
Very simple fix to #415 as proposed by @rndme

### DIFF
--- a/test/render-test.js
+++ b/test/render-test.js
@@ -5,7 +5,11 @@ var path = require('path');
 var _files = path.join(__dirname, '_files');
 
 function getContents(testName, ext) {
-  return fs.readFileSync(path.join(_files, testName + '.' + ext), 'utf8');
+  try {
+    return fs.readFileSync(path.join(_files, testName + '.' + ext), 'utf8');
+  } catch (ex) {
+    return null;
+  }
 }
 
 function getView(testName) {
--- a/.gitignore
+++ b/.gitignore
@@ -5,3 +5,4 @@ dojox
 yui3
 qooxdoo.mustache.js
 
+npm-debug.log
--- a/mustache.js
+++ b/mustache.js
@@ -481,7 +481,7 @@
       for (var j = 0, valueLength = value.length; j < valueLength; ++j) {
         buffer += this.renderTokens(token[4], context.push(value[j]), partials, originalTemplate);
       }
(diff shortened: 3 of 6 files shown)


dasilvacontin:
I'd like to understand why it was failing before, I feel like there is something else wrong out there. Sadly, I don't have time right now. Could this wait a couple of days?

phillipj:
Absolutely. I see what you mean... Maybe a sign that `_renderSection`  needs even more readability fixing

dasilvacontin:
I've seen some `token[4]`'s out there, etc, it doesn't help at all to someone not familiar with the codebase. But besides that, I sense some kind of bug with dot notation, but not sure right now.

dasilvacontin:
Alright, I've been able to take a look today, and see why it wasn't working before with my own eyes. :)

dasilvacontin:
@phillipj, could you add tests that assert `{{.}}` works for when any type of variable/object becomes the `Context`?

test/render-test.js
```diff
@@ -5,7 +5,11 @@ var path = require('path');
 var _files = path.join(__dirname, '_files');
 
 function getContents(testName, ext) {
-  return fs.readFileSync(path.join(_files, testName + '.' + ext), 'utf8');
+  try {
+    return fs.readFileSync(path.join(_files, testName + '.' + ext), 'utf8');
+  } catch (ex) {
+    return null;
+  }
```
phillipj: Included a small fix for running render tests which doesnt exist (e.g. misspelled), instead of some verbose lowlevel fs error.

Results in something like
```
Error: Cannot find view for test "mytest"
   ... <more stacktrace>
```
rather than
```
fs.js:439
  return binding.open(pathModule._makeLong(path), stringToFlags(flags), mode);
                 ^
Error: ENOENT, no such file or directory ....
```

phillipj:
@dasilvacontin just added a case for String as `Context`.

The third option `typeof value === 'object'` results in `[object Object]` getting rendered with `{{.}}`. I cannot think of a more appropriate value for that scenario, not too sure a test verifying that behavior is of too much value either.

test/render-test.js
dasilvacontin: Great!

dasilvacontin:
Fair enough, looks good! Thanks for the PR :)

Issue #442, opened 2015-04-10
TypeError in firefox if there is a tag named "watch".


Lukesed:
Minimal example:
Mustache.render("{{watch}}", {watch: "test"})

Result in Chrome 41:
"test"

Result in Firefox 37:
TypeError: missing argument 1 when calling function value.call

Perhaps it's caused by this: https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Object/watch

dasilvacontin:
Thanks for reporting! :)

phillipj:
I havent been able to reproduce this in firefox @Lukesed. Tried both locally with scripts and firebug + [browserstack](https://www.browserstack.com/).

How did you run your example too trigger that error? What exact firefox version and OS are you on?

Lukesed:
Error appears in the console when running this: https://jsfiddle.net/uno2xw56/

Firefox version 37.0.2, OS X 10.10.3.

Eschon:
Also appears on Firefox 39.0a2, OS X 10.10.3

phillipj:
Thanks guys! Confirmed the error after realizing I had a bug in my browser tests script. 

Bugfix coming soon :)

PR #443, opened 2015-04-12
Browser tests

phillipj:
This is WIP-branch to get the project covered by browser tests, not only V8 via node/iojs. It runs tests with [zuul](https://github.com/defunctzombie/zuul) which provides simple local browser testing, aswell as cloud testing on multiple browsers with [saucelabs](http://saucelabs.com).

**Running tests locally**
```bash
$ npm run test-browser-local
```

**Running in the cloud**
It requires a saucelabs account and [~/.zuulrc config](https://github.com/defunctzombie/zuul/wiki/Cloud-testing#2-educate-zuul):
```bash
$ npm run test-browser
```

**Issues**
Atm only `context-test.js` and `parse-test.js` are runned. The biggest test suite `render-test.js` still needs some creating thinking, as it reads test fixtures and expected output from files in `test/_files` which doesnt work too well in browsers.

Greatly appreciate any thoughts about this, especially how we'll get `render-test.js` running in browsers.

### DIFF
--- a/test/render-test.js
+++ b/test/render-test.js
@@ -1,72 +1,27 @@
 require('./helper');
 
-var fs = require('fs');
-var path = require('path');
-var _files = path.join(__dirname, '_files');
+var renderHelper = require('./render-helper');
 
-function getContents(testName, ext) {
-  try {
-    return fs.readFileSync(path.join(_files, testName + '.' + ext), 'utf8');
-  } catch (ex) {
-    return null;
-  }
-}
-
-function getView(testName) {
-  var view = getContents(testName, 'js');
-  if (!view) throw new Error('Cannot find view for test "' + testName + '"');
-  return eval(view);
-}
-
-function getPartial(testName) {
-  try {
-    return getContents(testName, 'partial');
-  } catch (error) {
-    // No big deal. Not all tests need to test partial support.
-  }
-}
-
-function getTest(testName) {
-  var test = {};
-  test.view = getView(testName);
-  test.template = getContents(testName, 'mustache');
-  test.partial = getPartial(testName);
-  test.expect = getContents(testName, 'txt');
-  return test;
-}
-
-// You can put the name of a specific test to run in the TEST environment
-// variable (e.g. TEST=backslashes vows test/render-test.js)
-var testToRun = process.env.TEST;
-
-var testNames;
-if (testToRun) {
-  testNames = [testToRun];
-} else {
-  testNames = fs.readdirSync(_files).filter(function (file) {
-    return (/\.js$/).test(file);
-  }).map(function (file) {
-    return path.basename(file).replace(/\.js$/, '');
-  });
-}
+var tests = renderHelper.getTests();
 
 describe('Mustache.render', function () {
   beforeEach(function () {
     Mustache.clearCache();
   });
 
-  testNames.forEach(function (testName) {
-    var test = getTest(testName);
+  tests.forEach(function (test) {
+    var view = eval(test.view);
 
-    it('knows how to render ' + testName, function () {
+    it('knows how to render ' + test.name, function () {
       var output;
       if (test.partial) {
-        output = Mustache.render(test.template, test.view, { partial: test.partial });
+        output = Mustache.render(test.template, view, { partial: test.partial });
       } else {
-        output = Mustache.render(test.template, test.view);
+        output = Mustache.render(test.template, view);
       }
 
       assert.equal(output, test.expect);
     });
+
   });
 });
--- a/.gitignore
+++ b/.gitignore
@@ -6,3 +6,5 @@ yui3
 qooxdoo.mustache.js
 
 npm-debug.log
+
+test/render-test-browser.js
--- a/.travis.yml
+++ b/.travis.yml
@@ -6,6 +6,13 @@ node_js:
   - 0.12
 before_install:
   - "test $TRAVIS_NODE_VERSION = '0.6' && npm install -g npm@1.3.26 || npm install -g npm"
+script:
+  - npm test
+  - npm run test-browser
 matrix:
   allow_failures:
     - node_js: 0.11
+env:
+  global:
+    - secure: TKDF1BgAR3S1bqFvikN6O3nbcLk20PDkL6QUVS6IkqQNtL1wXK4+2+GKcvJzOjCZg4OnXAFFtF/rfh9bcNlrGtWCG9CFkuGjUM7iONSQJr0kvCn7jris2jGGBupoPkX7MKlxaqrg6bmbp1QgA42/Hu4AlsNPCBiuFjpvL67VSqg=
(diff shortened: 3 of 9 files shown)


phillipj:
**Update**

Got tests from `render-tests.js` running aswell. In other words; no more work to be done.

Used mustache.js (ofcourse :confetti_ball:) too render all fixtures from `_tests` into one single test file which browsers can run. Primarly:

```bash
$ node test/create-browser-suite.js
# -> turns test/render-browser-tests.mustache into render-browser-test.js
```

As `render-browser-test.js` is a copy of all render tests, it is not commited into repo, but generated by the newly added npm scripts:

```bash
$ npm run browser-test
$ npm run browser-test-local
```

ATM these tests fail on Travis as Travis doesnt set encrypted $ENVs from pull requests for security reasons. Those variables are required by saucelabs too run tests. Although runned from my machine on saucelabs, shows that all 118 tests pass in looots of browsers :dancer:

Any thoughts and feedback?

phillipj:
@janl or @dasilvacontin do one of you have a couple of spare minutes? IMHO its alot better than zero browser tests anyhow... If I dont hear anything within a week, I'll solo merge this. 

I have already used this to confirm the reported firefox bug in #442, great success!

test/render-test.js
```diff
@@ -1,72 +1,27 @@
...
 
 describe('Mustache.render', function () {
   beforeEach(function () {
     Mustache.clearCache();
   });
 
-  testNames.forEach(function (testName) {
-    var test = getTest(testName);
+  tests.forEach(function (test) {
+    var view = eval(test.view);
```
dasilvacontin: Why not perform the `eval` inside the helper? Kinda looks a bit odd seeing it here.

Maybe you found an issue with giving the same view object to all the tests?

dasilvacontin:
> Browser tests are not included in `npm test` as they run for too long, although they are runned automatically on Travis upon commit.

> [...]

> ATM these tests fail on Travis as Travis doesnt set encrypted $ENVs from pull requests for security reasons. Those variables are required by saucelabs to run tests.

So this will currently break the Travis build?

phillipj:
> So this will currently break the Travis build?

Nope, as soon as this code is part of the `janl/mustache.js` repo, the encrypted $ENVs will be set correctly and tests runned on saucelabs. I just realized I should have created this feature branch in `janl/mustache.js` so that these encrypted $ENVs would have been set all along .. lesson learned.

dasilvacontin:
Oh, that's awesome!

Could you rebase the PR for a clean merge? Thanks. :)

test/render-test.js
phillipj: Agree that it looks kinda wierd.. The reason is because all the tests & fixtures are inlined in the browser suite generation by `render-test-browser-tmpl.mustache`. We need it to still be strings, not eval-ed code as we're inlining functions in some test views.

Could ofc hide it with some options provided to the helper when in node (e.g. `renderHelper.getTests({ evalViews: true })`), but we cannot get around it when testing browsers as far as I see.

dasilvacontin:
I've run out of testing minutes, but I think that there was some kind of issue with iOS 8.1, or at least it didn't pass the two/three times I tried running the tests.

However, the solution for browser testing works as expected! :+1:

phillipj:
Thanks for taking the time to check it out. About timeouts on saucelabs,
I've seen them occesionally aswell with iOS. Might be worth reporting to
saucelabs if it persists..

P.S. saucelabs has special accounts for open source projects / users.
Unlimited testing with 3 parallel runs 👍

test/render-test.js
phillipj: What do you think about `renderHelper.getTests({ evalViews: true })` or something similar @dasilvacontin ? Should I do a small refactoring with that change or is this PR good to go?

dasilvacontin: Hmm, it might be the 7UP, but I can't see why we can't `eval` objects that contain functions beforehand.

dasilvacontin: I'm taking a look at the code, maybe it's that I didn't look at the flow too much.

dasilvacontin: Yeah, I didn't read enough code. Makes sense now. :+1:

phillipj: Okey, lets say we do `eval` directly in `renderHelper.getView()`. For node it works perfectly :+1: Although for in browsers we now suddenly got 15 broken tests. Why is that? Example from 1 of the 15 broken tests: `./test/_files/escaped.js`:

```javascript
({
  title: function () {
    return "Bear > Shark";
  },
  entities: "&quot; \"'<>/"
})
```

`eval`ing that view in node is perfectly fine; it evaluates the view containing the function and all, ready to be invoked by the current process. Although for browsers its complete different because all the render tests, views and expected output has to be inlined ahead of time as browsers cant read the tests files we have in `./test/_files`. Browsers therefore need the raw views as string so they can `eval` them upon executing the tests on saucelabs which is a completely different process than the one reading `./test/_files`.

Did that clearify something at all? :smile:

phillipj:
I dig into the $ENVs right away ;)

test/render-test.js
dasilvacontin: I mainly didn't pay enough attention to the generator for browser tests (https://github.com/phillipj/mustache.js/commit/b473d19caec55dd111ed798e575217abf3db45fb#diff-1624925b7ed7bfb0de67878c2b18002dR12).

Thanks for explaining it though! :)

phillipj:
FYI I have gotten the tests to run and done some tweaking: 1a6396f4b06d00cb844d32a854da2eaaa70c9e24 

It takes ages to get Travis to rerun these tonight, so I'll probably have to continue this tomorrow.

phillipj:
Fixed in master with 35df920bd752357c28a9cfe7d4aebbc5230afed3

dasilvacontin:
Awesome! Why the `|| true` in `.travis.yml`, though?

phillipj:
Old habit of ensuring we're exiting successfully.

dasilvacontin:
But shouldn't it fail if the tests fail?

phillipj:
Absolutely. Looking at the test and command together:
```bash
$ test $TRAVIS_NODE_VERSION = '0.12' && npm run test-browser || true
```
1. Are we currently running node v0.12 on Travis? Then run the browser tests.
2. Obviously not v0.12, dont do anything but exit cleanly.

Why only on v0.12? 
- v0.8 doesnt work well with zuul because of some native addon code which isnt compatible with v0.8
- The tests are running in the browser, **not** node, therefore there is no reason to run the browser tests on node v0.10 also

dasilvacontin:
But what if it runs the tests and the tests fail? That's the case I was trying to point out.

dasilvacontin:
Shouldn't it be something like `node_version = 0.12 && npm run browser tests || !node_version = 0.12`? (pseudo code-ish)

dasilvacontin:
`node_version != 0.12 || npm run browser tests`

phillipj:
Ahhhh, I see what you mean. Really liked that last one, thanks :+1: I'll fix it shortly

dasilvacontin:
Cool! :)

PR #449, opened 2015-05-12
Minor changes to make it compatible with both WSH and ASP

nagaozen:
`<script>`, even in comments, triggers `Active Server Pages error 'ASP 0138', Nested Script Block` Error.

Also, using `global ← this` pattern works on WSH but not on ASP, it's better to just Mustache = {}, since its less typing, works on all platforms and it's the the desired behavior.

### DIFF
--- a/mustache.js
+++ b/mustache.js
@@ -11,7 +11,7 @@
   } else if (typeof define === 'function' && define.amd) {
     define(['exports'], factory); // AMD
   } else {
-    factory(global.Mustache = {}); // <script>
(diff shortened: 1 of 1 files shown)


dasilvacontin:
Hi @nagaozen, thanks for the PR!

Unfortunately, it's breaking the Travis build: the linter quits. Can you fix it? Cheers!

```
> eslint mustache.js
mustache.js
  14:12  error  "Mustache" is not defined  no-undef
```

phillipj:
Fixed with 60c2da80783a7

Issue #445, opened 2015-04-19
Problem with getter/setter accessors generated from TypeScript


wgebczyk:
For below HTML file, the alert output is: "[0;00]||[1;][3;]". Expected is: "[0;00]||[1;2][3;4]"

The problem lies in lines 383 & 388. The hasOwnProperty does not returns true.
If additionally we could check __proto__.hasOwnProperty, then this should work fine.

The HTML
```html
<!DOCTYPE html>
<html>
<head>
  <title>aaa</title>
  <script src="res/mustache.js"></script>
  <script>
    /* TypeScript src
      interface IAaa {
        x: string;
        y: string;
      }

      class Aaa implements IAaa {
        public x: string;
        private _y: string;

        constructor(x: string, y: string) {
          this.x = x;
          this._y = y;
        }

        get y(): string { return this._y; }
        set y(value: string) { this._y = value; }
      }

      class Bbb {
        item: IAaa;
        items: IAaa[];
      }
    */
    var Aaa = (function () {
      function Aaa(x, y) {
        this.x = x;
        this._y = y;
      }
      Object.defineProperty(Aaa.prototype, "y", {
        get: function () {
          return this._y;
        },
        set: function (value) {
          this._y = value;
        },
        enumerable: true,
        configurable: true
      });
      return Aaa;
    })();
    var Bbb = (function () {
      function Bbb() {
      }
      return Bbb;
    })();
  </script>
</head>
<body>
  <script>
    var b = new Bbb();
    b.item = new Aaa("0", "00");
    b.items = [];
    b.items.push({ a: new Aaa("1", "2") });
    b.items.push({ a: new Aaa("3", "4") });
    var output = Mustache.render('[{{ item.x }};{{ item.y }}]||{{#items}}[{{ a.x }};{{ a.y }}]{{/items}}', b);
    alert(output);
  </script>
</body>
</html>
```

dasilvacontin:
I confirm this bug.

phillipj:
Thanks! I'm investigating some prototype issues in the reported firefox
issue also. I'll take this into consideration aswell

wgebczyk:
tx!

PR #451, opened 2015-05-15
Fix view prototype value lookup issues

phillipj:
Fixes issue #445 where view object has properties defined in the view's prototype


dasilvacontin:
I think I found a bug. I'll push some unit tests that break.

phillipj:
Great!

dasilvacontin:
I added a failing test. I suggest rebasing onto `master` to get an improved `AssertionError` diff for the render tests.

phillipj:
Awesome @dasilvacontin, really appreciate your test!

phillipj:
FYI I'm splitting these fixes into two separate PRs, as the first fix context cache fix does not break the test you added

dasilvacontin:
Great! I already merged the other one, thanks! :)

test/render-helper.js
```diff
@@ -26,12 +26,12 @@ function getPartial(testName) {
...
 
 // You can put the name of a specific test to run in the TEST environment
-// variable (e.g. TEST=backslashes vows test/render-test.js)
+// variable (e.g. TEST=backslashes mocha test/render-test.js)
 var testToRun = process.env.TEST;
 
 var testNames;
 if (testToRun) {
-  testNames = [testToRun];
+  testNames = testToRun.split(',');
```
phillipj: Small enhancement enabling more than one explicit render test to be runned in isolation

phillipj:
Finally got around to fixing what caused your newly added test to break. Cleaned up the code abit by creating a re-usable method for checking property existence.

dasilvacontin:
It's awesome to have that check as a re-usable method. Awesome, thanks @phillipj!

Issue #429, opened 2015-02-26
Mustache.js template with function fails using command line tool


hhumml:
Using the examples from the readme section, I get the following exception when trying it with the command line tool (installed with npm):
```
undefined:3
	"bold": function () {
	         ^
SyntaxError: Unexpected token u
    at Object.parse (native)
    at /usr/local/lib/node_modules/mustache/bin/mustache:49:13
    at ReadStream.<anonymous> (/usr/local/lib/node_modules/mustache/bin/mustache:72:5)
    at ReadStream.g (events.js:180:16)
    at ReadStream.emit (events.js:117:20)
    at _stream_readable.js:944:16
    at process._tickCallback (node.js:442:13)
```
View
```
{
	"name": "Tater",
	"bold": function () {
		return function (text, render) {
			return "<b>" + render(text) + "</b>";
		}
	}
}
```
Template
```
{{#bold}}Hi {{name}}.{{/bold}}
```
When trying the same within the browser it works. What am I missing?

Background: I want to build a HTML template generator using the command line tool.

dasilvacontin:
Hi @hhumml, the view argument is supposed to be a .json file. JSON format doesn't support functions.

renancarvalhoo:
@hhumml have example in the documentation of mustache :
<pre>
Template:
<code>
{{#wrapped}}
  {{name}} is awesome.
{{/wrapped}}
</code>
Hash:
<code>
{
  "name": "Willy",
  "wrapped": function() {
    return function(text, render) {
      return "\<b>" + render(text) + "\</b>"
    }
  }
}
</code>
https://mustache.github.io/mustache.5.html

dasilvacontin:
@renancarvalhoo, you can't store functions in .json file.

renancarvalhoo:
i know this but you can work the data in the javascript before pass to template mustache.

dasilvacontin:
@renancarvalhoo, but @hhumml is talking about using the command line tool.

watersoona:
I have the same problem, how to solve it ?

PR #455, opened 2015-05-21
Improve CLI parsing error

phillipj:
I've become fund of human readable and informative error messages with tips lately ;) Would you agree this has some benefits compared to the raw exception stack from `JSON.parse()`?

Somewhat inspired by issue #429 expecting functions to be valid JSON.

```bash
→ echo {name:'lebron'} | ./bin/mustache - test/_files/cli.mustache 
undefined:1
{name:lebron}
 ^
SyntaxError: Unexpected token n
    at Object.parse (native)
    at /mustache.js/bin/mustache:49:13
```
vs
```bash
→ echo {name:'lebron'} | ./bin/mustache - test/_files/cli.mustache 
Shooot, could not parse view as JSON.
Tips: functions are not valid JSON and keys / values must be surround with double quotes.

SyntaxError: Unexpected token n
    at Object.parse (native)
    at parseView (/mustache.js/bin/mustache:55:17)
    at /mustache.js/bin/mustache:49:8
```

### DIFF
--- a/bin/mustache
+++ b/bin/mustache
@@ -46,8 +46,21 @@ function readView(cb) {
   var view = isStdin(viewArg) ? process.openStdin() : fs.createReadStream(viewArg);
 
   streamToStr(view, function(str) {
-    cb(JSON.parse(str));
+    cb(parseView(str));
   });
+}
+
+function parseView(str) {
+  try {
+    return JSON.parse(str);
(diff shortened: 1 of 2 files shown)


dasilvacontin:
This is awesome. Thanks @phillipj, love it! :smile:

PR #456, opened 2015-05-27
Added the keywords "imports", "nested" and "subtemplates" to the docs…

mmcev106:
… for Partials.  It took me forever to find this feature because those were the keywords I was looking for.  The term "partials" means something slightly different in some circles, so I kept misinterpretting their use when skimming the docs.

### DIFF
--- a/README.md
+++ b/README.md
@@ -382,7 +382,8 @@ Mustache requires only this:
 {{> next_more}}
 ```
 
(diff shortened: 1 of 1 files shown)


phillipj:
LGTM :+1:

phillipj:
Thanks!

PR #460, opened 2015-06-02
Only display result of master branch on travis badge

nikolas:


phillipj:
Nice, didnt know about that URL parameter. Thanks @nikolas!

PR #461, opened 2015-06-04
Bugfix for view context cache

### LINKED ISSUE #442: TypeError in firefox if there is a tag named "watch".

phillipj:
The context cache looked up previously cached values by matching `cache.property` rather than `cache.hasOwnProperty()`. This resulted in some values getting used from `Object.prototype` rather than ignoring that value.

Fixes #442

### DIFF
--- a/mustache.js
+++ b/mustache.js
@@ -356,7 +356,7 @@
     var cache = this.cache;
 
     var value;
-    if (name in cache) {
+    if (cache.hasOwnProperty(name)) {
       value = cache[name];
     } else {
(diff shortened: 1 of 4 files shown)


dasilvacontin:
LGTM!

PR #463, opened 2015-06-10
Remove moot `version` property from bower.json

kkirsche:
Per bower/bower.json-spec@a325da3

### DIFF
--- a/bower.json
+++ b/bower.json
@@ -1,7 +1,6 @@
 {
   "name": "mustache.js",
   "main": "mustache.js",
-  "version": "2.1.0",
   "homepage": "https://github.com/janl/mustache.js",
   "authors": [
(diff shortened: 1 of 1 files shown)


dasilvacontin:
LGTM, thanks for the link too!

kkirsche:
Yep. Also just an FYI there are discussions to move bower to NPM anyway so I wouldn't really plan to support it for ever: bower/bower#1520

kkirsche:
Also their maintainer says they probably won't ever use it in case people bug you about removing it: http://stackoverflow.com/questions/24844901/bowers-bower-json-file-version-property

dasilvacontin:
Thanks for the additional info. I don't need bower myself, nor I'm a fan/user.

PR #466, opened 2015-06-17
Added global Mustache definition

yousefcisco:
The Mustache variable isn't defined globally which means the code doesn't work correctly when used in strict mode.

### DIFF
--- a/mustache.js
+++ b/mustache.js
@@ -11,7 +11,7 @@
   } else if (typeof define === 'function' && define.amd) {
     define(['exports'], factory); // AMD
   } else {
-    Mustache = {};
+    global.Mustache = {};
     factory(Mustache); // script, wsh, asp
   }
 }(this, function mustacheFactory (mustache) {


bobthecow:
That was just removed by @nagaozen in 60c2da80783a71409a8c3ecbc32a3d4e858df691. Y'all are going to have to fight it out ;)

yousefcisco:
It looks like the removal was so it would be compatible with eslint, this change thankfully doesn't cause any eslint errors and allows the code to be run in strict mode.

The reason behind this PR is because we bundle Mustache.js with a bunch of other libraries that we use on our sites and automatically add `"use strict"` to the top of the merged script so this was causing it all to fail.

dasilvacontin:
Thanks for the commit reference @bobthecow, made it easy for figuring out @nagaozen's change was just for `eslint` compatibility.

I confirm this change doesn't raise any `eslint` errors.

@yousefcisco, thanks for the PR! I merge and release right now. :shipit:

phillipj:
https://github.com/janl/mustache.js/pull/449 was the original source for the global being removed, not just to satisfy eslint

yousefcisco:
What causes the issue in ASP? Surely this would also be an issue with other libraries as this pattern is widely used.

dasilvacontin:
Great, I misunderstood the commit message. I should have read the issue too.

I'm trying to figure out why the `global` way doesn't work in ASP. If you find anything, please post it here. /cc @nagaozen

PR #467, opened 2015-06-24
Throw error when providing .render() with invalid template type

### LINKED ISSUE #464: Add error handling when view is not an object in render()
I got stuck with this error for a long while "Uncaught TypeError: this.tail.search is not a function" and in the end somebody helped by pointing out I was passing render() an array rather than object for view.
I know a lot of people won't have issues with this, but as I explained in my post  http://morenafiore.com/2-things-i-did-not-understand-about-mustache/ it might help people who approach Mustache.js for the first time to not give up by having more specific errors displaying.

Other people had this error, but it was down to template not being passed as plain text  http://stackoverflow.com/questions/10872615/what-is-the-difference-between-mustache-render-and-mustache-to-html

I hope you might be able to help :)

phillipj:
As we require the template to be a `string` we now throw an error when given any other data type. This should provide the developers with a meaningful error rather than a cryptic TypeError from the murky depths of our source code.

Fixes #464

### DIFF
--- a/mustache.js
+++ b/mustache.js
@@ -25,6 +25,14 @@
     return typeof object === 'function';
   }
 
+  /**
+   * More correct typeof string handling array
+   * which normally returns typeof 'object'
+   */
+  function typeStr (obj) {
+    return isArray(obj) ? 'array' : typeof obj;
+  }
+
   function escapeRegExp (string) {
     return string.replace(/[\-\[\]{}()*+?.,\\\^$|#\s]/g, '\\$&');
   }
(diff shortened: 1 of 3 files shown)


dasilvacontin:
I'd like to give more info to the users, so they don't need to go back to the docs in order to fix the error. They might be confused about argument position besides argument type.

I usually mention the position of the argument we are talking about, but it's probably even better to state the function definition/parameters:

> Error: Invalid template! Template should be a "string" but "number" was given as the first argument for mustache#render(template, view, partials).

Or something similar.

phillipj:
Absolutely an improvement to display method signature :+1:

dasilvacontin:
```
> Mustache.render([])
TypeError: Invalid template! Template should be a "string" but "array" was given as the first argument for mustache#render(template, view, partials)
    at Object.render (/Users/dasilvacontin/temp/node_modules/mustache/mustache.js:596:13)
```

Beautiful, thanks @phillipj!

PR #469, opened 2015-07-16
Documents use of string literals containing double quotes

jfmercer:
Currently, the documentation is not clear about how to write a double-quoted word as a string literal. This PR corrects that. For example, if you want `{{name}}` to be interpreted as a string literal rather than a mustache tag, you must change and then restore the default delimiter. 

View:

```json
{
  "name": "literal"
}
```

Template:

```html
* {{name}}
{{=<% %>=}}
* {{name}}
<%={{ }}=%>
```

Output:

```html
* literal
* {{name}}
```

### DIFF
--- a/README.md
+++ b/README.md
@@ -106,6 +106,8 @@ The most basic tag type is a simple variable. A `{{name}}` tag renders the value
 
 All variables are HTML-escaped by default. If you want to render unescaped HTML, use the triple mustache: `{{{name}}}`. You can also use `&` to unescape a variable.
 
+To print `{{name}}` as a string literal, the default delimiters must be changed. See the ["Set Delimiter'](https://github.com/janl/mustache.js#set-delimiter) section for more information about custom delimiters.
+
 View:
 
 ```json
@@ -123,6 +125,9 @@ Template:
 * {{company}}
 * {{{company}}}
 * {{&company}}
+{{=<% %>=}}
+* {{company}}
+<%={{ }}=%>
 ```
 
 Output:
@@ -133,6 +138,7 @@ Output:
 * &lt;b&gt;GitHub&lt;/b&gt;
 * <b>GitHub</b>
 * <b>GitHub</b>
+* {{company}}
 ```
 
 JavaScript's dot notation may be used to access keys that are properties of objects in a view.


phillipj:
Thanks for improving the docs @jfmercer :+1:

README.md
```diff
@@ -106,6 +106,8 @@ The most basic tag type is a simple variable. A `{{name}}` tag renders the value
 
 All variables are HTML-escaped by default. If you want to render unescaped HTML, use the triple mustache: `{{{name}}}`. You can also use `&` to unescape a variable.
 
+To print `{{name}}` as a string literal, the default delimiters must be changed. See the ["Set Delimiter'](https://github.com/janl/mustache.js#set-delimiter) section for more information about custom delimiters.
+
```
bobthecow: I wouldn't use the phrase "string literal". A string literal is literally `"something like this"`, which isn't exactly what this means here. I'd say something like:

> If you want to mustache tags — like `{{name}}` — literally in your template, the default delimiters must be changed. See …

dasilvacontin: Yeah, `"{{ name }}"` and `"Bob"` are both string literals.. "Literally in your template" doesn't explain it clearly either, imo. What about mentioning "render"/"output"?

> If you want to render mustache tags – like `{{ name }}` – literally in your output...

bobthecow: That's not super clear either, because "render mustache tags" means "turn them into not mustache tags" :)

Romanx: Why not use something combining both?

> If you want to output mustache tags – like {{ name }} – in your output...

Although output followed by "in your output" could look a little odd

jfmercer: I'm happy with the language as-is. However, if you guys would prefer something else, let me know what you settle on, & I'll submit a PR.

dasilvacontin: > That's not super clear either, because "render mustache tags" means "turn them into not mustache tags" :)

\>.<

jfmercer: I believe that @bobthecow is right in objecting to the use of "string literal" in this case. I'd suggest avoiding the use of the word "literal" altogether because (1) "literal" has special significance in computer language jargon, which isn't necessary here, and (2) unfortunately, in modern English the words "literal" and "literally" often mean their opposites, "figurative" and "figuratively." 

How about this?

If you want `{{name}}` *not* to be interpreted as a mustache tag and to appear simply as `{{name}}` in the output, you must change and then restore the default delimiter.

phillipj: +1 for "output" if string literal needs to change

dasilvacontin: I like what you suggested @jfmercer. Thinking of ways to improve `and to appear simply as {{name}}` though.

jfmercer: I'll submit a new PR for review by Sunday evening.

jfmercer:
I have submitted an update in PR #474. Discussion and suggestions are welcome.

PR #472, opened 2015-07-22
Update UMD Shim to be resilient to HTMLElement global pollution

mikesherov:
Please see https://github.com/umdjs/umd/pull/63 and https://github.com/umdjs/umd/issues/35 for more information, but this change prevents an HTML element with an id of `exports` from causing the UMD shim to incorrectly detect the environment as node.

### DIFF
--- a/mustache.js
+++ b/mustache.js
@@ -6,7 +6,7 @@
 /*global define: false Mustache: true*/
 
 (function defineMustache (global, factory) {
-  if (typeof exports === 'object' && exports) {
+  if (typeof exports === 'object' && exports && typeof exports.nodeName !== 'string') {
     factory(exports); // CommonJS
   } else if (typeof define === 'function' && define.amd) {
     define(['exports'], factory); // AMD


phillipj:
LGTM

dasilvacontin:
@mikesherov could you shed some light on why a null check is not necessary, or at least in your opinion?

mikesherov:
@dasilvacontin that's interesting. I suppose technically it's needed, but only if bad actors are setting `exports` to `null`. In the case of the nodeName check, it addresses the real possibility of accidental footguns.

I'm not opposed to adding back in the truthy check. I'll amend.

mikesherov:
@dasilvacontin updated.

dasilvacontin:
Indeed, that was the concern. Although rarely, I can see that situation happening to some people, specially newcomers.

Thanks, I'll merge. :)

mikesherov:
@dasilvacontin thanks for merging! Any chance we can get a patch release?

dasilvacontin:
I'll do so asap, probably in less than an hour. :shipit:

mikesherov:
@dasilvacontin did something prevent a bump here?

dasilvacontin:
Oups. Yeah, I had to set up a new env, and then I got caught up with work. Will do now, thanks for the ping. :)

mikesherov:
@dasilvacontin do you need to push the tag up to the repo for it to appear in bower?

dasilvacontin:
Yes, I do. Sorry, it's been a while, and I had forgotten about bower. I've also been meaning to update the bump script...

mikesherov:
No worries! Sorry to keep bothering you, and thanks again for the quick attention!

dasilvacontin:
No problemo, you're welcome. :)

PR #473, opened 2015-07-23
Inheritable Blocks Sections

damyon:
This is a patch for a feature that is not in the spec, but is proposed here:

https://github.com/mustache/spec/pull/75

It is implemented in at least GRMustache, mustache.php and hogan.js.

See mustache.php docs here: https://github.com/bobthecow/mustache.php/wiki/BLOCKS-pragma (note the caveat about lazy evaluation on that page has been fixed).

The tests in this pull request mostly come from GRMustache, but I couldn't add all his tests, because some require multiple partials.


damyon:
Note - I don't mind if this is or isn't merged as it looks like you are only accepting things that are already agreed to and in the spec - I just wanted create this pull request in case someone else is looking for this feature.

groue:
FYI, [GRMustache.swift](https://github.com/groue/GRMustache.swift) has a cleaner implementation than its Objective-C ancestor, and passes [more tests](https://github.com/groue/GRMustache.swift/blob/master/MustacheTests/Public/SuitesTests/GRMustacheSuite/GRMustacheSuite/inheritable_partials.json). I don't know if those extra tests involve the "multiple partials" you refer to.

As stated in https://github.com/mustache/spec/issues/38#issuecomment-120612194, GRMustache.swift implementation and tests have a simple goal: any part of a template can be refactored with partials and inherited partials, without requiring any modification anywhere else (in other templates that depend on it, or in application code).

damyon:
My comment about not including some tests was only because of the way unit tests are created for this mustache.js. These tests are only ever expecting a single file named blah.partial - which is added to the config as a partial named 'partial' in the test. So tests requiring more than one partial cannot be executed by this test suite.

ChesterOfTheSemester:
Is this still supported?

PR #487, opened 2015-09-22
Updated eslint to version 1.5.1

janl:
ignore


phillipj:
Should we use version ranges instead of continuously bumping version numbers?

janl:
sorry, this is testing something, ignore

PR #488, opened 2015-09-24
Pinned all dependencies

janl:
:wave: Hello!

We’re all trying to keep our software up to date, yet stable at the same time.

This is the first in a series of automatic PRs to help you achieve this goal. It pins all of the dependencies in your `package.json`, so you have complete control over the exact state of your software.

From now on you’ll receive PRs whenever one of your dependencies updates – in real time. This way you get all the benefits of up-to-date dependencies, while having them pinned at the same time. This means:

- No more surprises because some of your dependencies didn’t follow [SemVer](http://semver.org/). Your software is always in a state you know about, regardless of _when_ someone does `$ npm install`.
- If you have continuous integration set up for this repo, it’ll run automatically. Ideally, it will pass and you'll stay up to date with the press of a button. If the updated dependency _does_ break your software, it won’t affect your users, because their dependencies are still pinned to the working state.
- You can immediately identify which dependency updates break your software, because each one is tested in isolation. You’ll also have a branch ready to work on, so adapting to new APIs is way faster.

Merge this PR and you’ll have up-to-date software without the headaches :sparkles:

---

This pull request was created by [greenkeeper.io](http://greenkeeper.io/).
It keeps your software, up to date, all the time.

<sub>
Tired of seeing this sponsor message? Upgrade to the supporter plan!
You'll also get your pull requests faster :zap:
</sub>

### DIFF
--- a/package.json
+++ b/package.json
@@ -42,7 +42,7 @@
   "devDependencies": {
     "chai": "3.3.0",
     "eslint": "1.5.0",
-    "mocha": "^2.1.0",
+    "mocha": "2.3.3",
     "zuul": "^2.1.1"
   },
   "spm": {


phillipj:
Yo @janl! Knowing this also might be a test PR, I would like challenge you on pinning and whether or not it provides complete control of the dependency tree.

Even though we pin a version of a dependency, those dependencies (and their deps again) very often have version ranges which results in dependency trees not being 100% reproduceable over time. Would you agree or have I misunderstood your meaning with complete control here?

janl:
@phillipj yeah, still testing, sorry :)

The plan would be to get all of npm on this model, eventually. We have to start somewhere.

phillipj:
Oooh, so on the quest of killing version ranges and semver are you? :smiley_cat: Understandable when having wrestled with `npm shrinkwrap` more than once.. Although I must say I have had very little issues with module authors breaking the laws of semver, maybe I have been lucky with the modules I have chosen in the past.

janl:
yeah, why trust humans, when robots can do the job much more accurately ;)

https://github.com/semantic-release/semantic-release

PR #490, opened 2015-10-02
Fix #489: Section functions not respecting modified delimiters

### LINKED ISSUE #489: Always need to set delimiters in custom render(text) function blocks
I have a template code

```
{{=<. .>=}}
    <.#post.>
    <div class="post <.type.>">
        <.#show_high_res.>
            <.#photoset_images.>
                <.#highres.>
                    <.#insert_image.>
                        <.&permalink.>
                    <./insert_image.>
                <./highres.>
            <./photoset_images.>
        <.#show_high_res.>
    </div>
    <./post.>
```
And it outputs `<.&permalink.>` as <.&permalink.> - it doesn't change the variable to a value.

The I noticed if I write `{{permalink}}` instead, it changes the variable with the value assigned in the JS object.

```
{{=<. .>=}}
    <.#post.>
    <div class="post <.type.>">
        <.#show_high_res.>
            <.#photoset_images.>
                <.#highres.>
                    <.#insert_image.>
                        {{&permalink}}
                    <./insert_image.>
                <./highres.>
            <./photoset_images.>
        <.#show_high_res.>
    </div>
    <./post.>
```

Then I tried setting the delimiter inside `<.insert_image.>` block and it finally rendered with a value instead of the variable name.

```
{{=<. .>=}}
    <.#post.>
    <div class="post <.type.>">
        <.#show_high_res.>
            <.#photoset_images.>
                <.#highres.>
                    <.#insert_image.>
                        {{=<. .>=}}
                        <.&permalink.>
                    <./insert_image.>
                <./highres.>
            <./photoset_images.>
        <.#show_high_res.>
    </div>
    <./post.>
```

Is this normal behavior or is it a bug?

vgel:
ref: #489

Because of the way mustache currently treats delimiter setting tags (e.g. `{{=<% %>=}}`), and how section functions currently pull substrings directly from the original template source, delimiters set outside of a section definition would not be respected inside that section. This PR should fix this issue by passing the `tags` value through `Writer.render` and its family of rendering functions. As a bonus, it also makes it possible to set `tags` without an actual tag in the template source, like so:

```
(new m.Writer).render('<%val%> world', {val: 'hello'}, null, ['<%', '%>']) // -> 'hello world'
```


deadsource:
So I've seen that the README.md was updated a few hours ago. Why is this not getting any attention? Does this bug fix need additional testing? Improving? Maybe something else?

dasilvacontin:
@deadsource Needs reviewing by maintainers, hopefully someone has some spare time soon. Looks promising.

phillipj:
Sorry for a very late review here @vgel! This indeed looks very promising.

The only thing which comes to mind, is whether or not this could be considered a breaking change for partials.
Previously delimiter tag settings didn't affect to partials at all, now they do - sort of at least, I got some strange results while testing it.

Here's what I did to put partials into the mix aswell:

**test/_files/section_functions_delimiters.mustache**
```
{{=<% %>=}}
<%#func%><%value%><%/func%>
<%>partial%>
```

**test/_files/section_functions_delimiters.partial**
```
{{#func}}<%value%>{{/func}}
```

**test/_files/section_functions_delimiters.txt**
```
!hello world
!hello world
```

Were these changes meant to have any effect on partials?

vgel:
Hmm, that's strange. Is the partial using both sets of delimiters? I'll fix that, I think it makes more sense to set the delimiters on a per-file basis.

bobthecow:
Set delimiter tags should not affect partials. If the mustache spec doesn't cover that, it should :)

Romanx:
@bobthecow I had to actually go check my C# lib implementation but that was my understanding as well. Partials tag contexts get defaulted but can be overridden internally if required.

jamesaspence:
Is this going to get merged in? Right now I have to resort to adding my custom delimiter in the callback for the function, like so: `return render('{{=@{ }=}}' + value);`

PR #495, opened 2015-10-06
chore(package): pinned dependencies

adam-gilman:
http://greenkeeper.io/


phillipj:
Not able to merge this because of conflicts, could you have a look at it?

dasilvacontin:
greenkeeper.io will already ping us if a version inside the range breaks.

PR #497, opened 2015-10-07
Remove IE8 from zuul test runs

phillipj:
Running tests in IE8 stopped working backing in may when chai got introduced for assertions. It doesnt fail, but it shows as a red warning in the zuul output.

`- failed: <internet explorer 8 on Windows 2003> (0 failed, 0 passed)`

As IE8 gets less and less usage and support around the interwebs, I vote for running tests in IE9 =&gt;

*Commit introducing chai with comments:*  https://github.com/janl/mustache.js/commit/954a8fbf15dc956b754355e9a0681202628724b3

### DIFF
--- a/.zuul.yml
+++ b/.zuul.yml
@@ -5,4 +5,4 @@ browsers:
   - name: firefox
     version: latest
   - name: ie
-    version: 8..latest
+    version: 9..latest


phillipj:
Got any objections to merging this @dasilvacontin?

dasilvacontin:
I don't mind dropping support for IE8 when building a project. But dropping IE8 support in a library...

You are kinda forcing everyone to drop support for that 'platform', or to not use your library... Chai did that, apparently.

I don't mind dropping support for IE8 and that should happen someday... But I feel bad for devs who are asked to support IE8...

On the other hand, giving them arguments like "hey, we barely find libraries that support IE8, or find none to do this task" might make corporate people want to move away from IE, upgrade their systems, which would be good...

What are your thoughts? What are @janl's thoughts?  I don't find myself experienced on this topic. :)

janl:
Could we have one last good branch that support IE8? Then we can point people there if they REALLY REALLY need it, but they are on their own with any new features or security updates. Then master can be IE9 and later.

phillipj:
> Could we have one last good branch that support IE8?

Sure! By not using chai for assertions, our current version passes all tests in IE8. We could publish a special version, e.g. `v2.2.0-ie8`?

dasilvacontin:
@phillipj publish a special version? Chai is only used for the tests, or am I missing something?

But yeah, we should keep a branch with chai-less tests for IE8 testing.

phillipj:
@dasilvacontin nope, you're right. Just thought we could push a last version proven to have passed all the tests in IE8.

If you meant a git branch specific to IE8 which doesnt use chai, what's the main purpose of that branch?

dasilvacontin:
@phillipj oh, I see.

We would use that branch to test the current version in IE8. But nevermind about it, we release a last good IE8 version, and we forget about it's existence... :smile:

dasilvacontin:
:ship:

PR #500, opened 2015-10-11
Removed html syntax highlighting on README.md example 

imagentleman:
This is just a silly request/nitpick, but with the html syntax highlighting this example looks just like plain text, except for the ampersand, which appears highlighted in red. This could be cause confusion, signaling that there is something "more important" about the ampersand than the rest of keywords (e.g. the delimiters).

### DIFF
--- a/README.md
+++ b/README.md
@@ -119,7 +119,7 @@ View:
 
 Template:
 
-```html
+```
 * {{name}}
 * {{age}}
 * {{company}}


dasilvacontin:
@imagentleman, I agree with you. It's a shame that GitHub (using [Linguist]) doesn't recognise Mustache as a language and doesn't support syntax highlighting for it (but it does recognise Handlebars).

[Linguist]: https://github.com/github/linguist

dasilvacontin:
Could you amend the commit's message to something more descriptive? Like `Remove syntax highlighting in README code sample`.

imagentleman:
@dasilvacontin, just amended the commit message : )

bobthecow:
I've had better luck with `html+jinja` for mustache snippets. Wanna give that one a shot here?

imagentleman:
@bobthecow Unfortunately, for this snippet in particular html+jinja doesn't highlight anything (does the same as the empty fence block in the pr).

phillipj:
Thanks!

PR #503, opened 2015-10-15
Linting CLI tool & fixing its current style issues

phillipj:
Ensure the same code style on the CLI tool, as we do on the main file `./mustache.js`.

Also minor .eslintrc change to be able to use `process.exit()`, instead of having to `throw new Error()` to exit the process.


dasilvacontin:
Cool, thanks! :rocket:

PR #510, opened 2015-10-20
fix indent error in switch-case statement

norfish:
When I read the code, I find there's an  indent error in  in switch-case statement.


dasilvacontin:
Thanks @norfish! :)

PR #524, opened 2015-10-30
Unpin chai and eslint versions

dasilvacontin:
greenkeeper.io will let us know if the minor/patch bumps break our code, so go wild.


phillipj:
👍

PR #526, opened 2015-10-31
added comments and I think I found a bug

kyl3r92:
New to Github, very good organized project and funny Idea! 
I just added some comments (it is already well commented) and I think I found a bug. 
Please review

### DIFF
--- a/mustache.js
+++ b/mustache.js
@@ -57,6 +57,7 @@
     return !testRegExp(nonSpaceRe, string);
   }
 
+  // Urlencoding / escaping map
   var entityMap = {
     '&': '&amp;',
     '<': '&lt;',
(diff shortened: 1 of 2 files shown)


mustache.js
```diff
@@ -65,13 +66,14 @@
...
-
+  
+  // Escapes urls
   function escapeHtml (string) {
     return String(string).replace(/[&<>"'\/]/g, function fromEntityMap (s) {
       return entityMap[s];
     });
   }
-
+
```
102: Remove extra space symbols from empty line

wrappers/jquery/mustache.js.pre
```diff
@@ -5,5 +5,6 @@ Shameless port of a shameless port
 See http://github.com/defunkt/mustache for more info.
 */
  
+ // This seems useless and not valid js syntax @TODO
```
dasilvacontin: It's actually correct js syntax, and not useless – it's part of the jquery wrapper generator, the header part.

dasilvacontin:
The comments don't really provide additional value, they are practically repeating the name of the variable.

PR #529, opened 2015-11-10
Sanitizer callback for unescaped values

dvoytenko:
/cc @cramforce
/cc @janl

### DIFF
--- a/mustache.js
+++ b/mustache.js
@@ -550,8 +550,12 @@
 
   Writer.prototype.unescapedValue = function unescapedValue (token, context) {
     var value = context.lookup(token[1]);
-    if (value != null)
+    if (value != null) {
+      if (mustache.sanitizeUnescaped) {
+        return mustache.sanitizeUnescaped(value);
+      }
       return value;
+    }
   };
 
   Writer.prototype.escapedValue = function escapedValue (token, context) {
@@ -618,6 +622,12 @@
   // Export the escaping function so that the user may override it.
   // See https://github.com/janl/mustache.js/issues/244
   mustache.escape = escapeHtml;
+
+  // Export the sanitizing function for unescaped values.
+  mustache.sanitizeUnescaped = null;
+  mustache.setUnescapedSanitizier = function setUnescapedSanitizier (sanitizeUnescaped) {
+    mustache.sanitizeUnescaped = sanitizeUnescaped;
+  };
 
   // Export these mainly for testing, but also for advanced usage.
   mustache.Scanner = Scanner;
--- a/test/sanitize-test.js
+++ b/test/sanitize-test.js
@@ -1,0 +1,18 @@
+require('./helper');
+
+var renderHelper = require('./render-helper');
+
+var tests = renderHelper.getTests();
+
+describe('Mustache.sanitizeUnescaped', function () {
+  beforeEach(function () {
+    Mustache.setUnescapedSanitizier(function(value) {
+      return value.toUpperCase();
+    });
+  });
+
+  it('requires template to be a string', function () {
+    assert.equal(Mustache.render('{{{value}}}', {value: '<b>abc</b>'}),
+        '<B>ABC</B>');
+  });
+});


phillipj:
You dont have to change mustache.js core to do this.

```js
var Writer = Mustache.Writer;
var _unescapedValue = Writer.prototype.unescapedValue

Writer.prototype.unescapedValue = function (token, context) {
    var value = _unescapedValue(token, context);
    return value ? value.toUpperCase() : undefined;
};

console.log(Mustache.render('{{{value}}}', {value: '<b>abc</b>'}));
```
https://jsfiddle.net/777e9s3n/

dasilvacontin:
However, it's not part of our API, and it's subject to change (though not a very likely change).

You could have tests and, in case we change our internals, @greenkeeperio-bot would notify you in case your plugin is no longer compatible.

dasilvacontin:
Here's an example of a plugin: https://github.com/dasilvacontin/mustache-i18n.

dvoytenko:
Yes, Writer override  is possible. However, I was following the implementation of `escapeHtml` just one line above. This change seemed more inline with that approach. However, it's more important since it involves a possibility of XSS.

phillipj:
> I was following the implementation of escapeHtml just one line above.

Ahhh, I see. This is seems like a very rare use case, so I'm kind of split about adding explicit methods to enable it, when one could get the same result with "advanced usage".

Thinking out loud, this boils down to two alternatives for me:

1. Provide a better way to override the internal `defaultWriter` variable, making Writer overriding less hackie
2. Refactor/simplify your `unescapeSanitizer` to look more like `mustache.escape`. Always set a default sanitizer, remove `setUnescapeSanitizer()` in favor of just setting the `mustache.unescapeSanitizer` property directly and always call `mustache.unescapeSanitizer()` in `Writer.proto.unescapeValue`

Any thoughts about these two approaches?

dvoytenko:
@phillipj I like the idea of 2. The code can definitely be made to look the same as "escapeHtml" above. One reason I added the setter is because export properties are a bit harder to work with in ES6. But not impossible, so I'd be happy to change.

One thing I'd disagree with is this being a rare case. It should be a very hard sell to allow an unescaped HTML output from data source to be done unfiltered. This is a way too easy to get an XSS. E.g.

Template:
```
{{{value}}}
```

Data:
```
{
  value: '<style>body {background: black}</style>'
}
```

Or worse.

phillipj:
> One thing I'd disagree with is this being a rare case. It should be a very hard sell to allow an unescaped HTML output from data source to be done unfiltered.

I dont follow your reasoning here. `{{value}}` would escape HTML, thereby avoid XSS. What does your unescape sanitizer in combination with `{{{value}}}` do in that respect?

dvoytenko:
@phillipj 

> What does your unescape sanitizer in combination with {{{value}}} do in that respect?

A sanitizer here would either strip any markup or would whitelist few tags. It's rather typical to allow basic formatting elements `<b>`, `<i>`, `<em>` and such.

phillipj:
> A sanitizer here would either strip any markup or would whitelist few tags.

Would you be able to do the same with `{{value}}` and a custom `mustache.escape()` method?

dvoytenko:
Sorry for delay.

> Would you be able to do the same with {{value}} and a custom mustache.escape() method?

Not really. `{{var}}` has a specific meaning that I'd not want to change. Do you have some resistance to an option to sanitize triple-mustache? This seems like a very common requirement. We want to allow unescaped HTML, but we don't want anyone to try to inject `<script></script>` this way. On the other hand `<b></b>` are totally fine.

phillipj:
>  Do you have some resistance to an option to sanitize triple-mustache?

Not against sanitize triple-mustache in particaluar, no. What I'm very conscious is unnecessary additions to core. We're the ones having to support this going forward. That's why I'm trying my best to understand what you're actually trying to accomplish.

> This seems like a very common requirement.

Has any other mustache implementation seen the need for this? Can't find any other issue in the repo asking for something similar in its 6 years of existence.

My thoughts are this doesn't belong in core. Why? Because (1) `{{{text}}}` or `{{& text}}` are meant for outputting the value as is, as simple as that. And (2) you could easily do sanitize the text before invoking `mustache.render()`.

**Sanitizer in mustache**
```js
const mustache = require('mustache');

mustache.sanitizeUnescaped = removeUnwantedValues;

const template = 'Text: <p>{{{content}}}</p>';
const view = {
  content: 'Not to be <strong>escaped</strong>'
};

console.log(mustache.render(template, view));
```

**Explicit sanitizing**
```js
const mustache = require('mustache');

const template = 'Text: <p>{{{content}}}</p>';
const view = {
  content: removeUnwantedValues('Not to be <strong>escaped</strong>')
};

console.log(mustache.render(template, view));
```

The latter has the advantages of being more explicit and less complex, which I'm in favor of.

dasilvacontin:
@dvoytenko IMO sanitation should be made before storing the text in a database / wherever, not during rendering – you want to provide same behaviour across platforms. You could also do it when preparing the view, if really necessary.

Closing due to lack of activity, and since it doesn't really look like it should go into core. If you think otherwise, reply to our feedback.

PR #531, opened 2015-11-23
Fixed double escaping

Silviu-Marian:
calling `escapeHtml(escapeHtml(escapeHtml(escapeHtml('&'))))` results in
`&amp;amp;amp;amp;`

### DIFF
--- a/mustache.js
+++ b/mustache.js
@@ -58,7 +58,6 @@
   }
 
   var entityMap = {
-    '&': '&amp;',
     '<': '&lt;',
     '>': '&gt;',
     '"': '&quot;',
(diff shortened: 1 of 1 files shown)


dasilvacontin:
The behaviour pointed out as erroneous is actually intended.

e.g. intended behaviour
```js
escapeHtml('In HTML, & should be escaped as &amp;.')
// => "In HTML, &amp; should be escaped as &amp;amp;."
```

whereas you suggest:
```js
escapeHtml('In HTML, & should be escaped as &amp;.')
// => "In HTML, &amp; should be escaped as &amp;."
```

which is not correct.

If this is a problem for you, you probably should know/control which arguments or variables in your code contain an already escaped string.

Silviu-Marian:
That's the problem, I'm taking data from a store and saving it back later. 

I did the pull request because I thought it's far more unlikely for a user to have to write specifically `&amp;`s than do it deliberately. We use &s in more frequently than we teach people HTML entities. 

And besides, it's trickier to turn `&amp;amp;` back to an `&amp;` than just avoid converting it. 

Would be nice to have at least `escapeHtmlOnce()` as a function.

phillipj:
> That's the problem, I'm taking data from a store and saving it back later.

There's lots of resources out there about tackling this, whether or not to store HTML encoded or not and where in the stack to encode.

> Would be nice to have at least escapeHtmlOnce() as a function.

Sounds like a new npm module :)

PR #532, opened 2015-11-26
Fix inconsistency in defining global mustache object.

simast:
A small fix as `global.Mustache` and `Mustache` may refer to two different things at this point. It worked before ok when Mustache was included in the default global `window` context - but for example if the script would have been included by wrapping Mustache code inside a function - this did cause issues.

### DIFF
--- a/mustache.js
+++ b/mustache.js
@@ -12,7 +12,7 @@
     define(['exports'], factory); // AMD
   } else {
     global.Mustache = {};
-    factory(Mustache); // script, wsh, asp
+    factory(global.Mustache); // script, wsh, asp
   }
 }(this, function mustacheFactory (mustache) {
 


phillipj:
Thanks for your contribution @simast! It would be great to see your issue with a small code example, both for us to understand it and for reference in the future.

simast:
Hey @phillipj!

This fix solves the problem of including Mustache script in the browser without adding a global Mustache instance as part of `window` context. The reason for doing this would be to not pollute the global namespace - or to have multiple non-conflicting Mustache objects (even at different version levels). I want to point out that this did work fine in an older Mustache version (1.x) and seems to have been introduced as a regression recently (probably with this commit: a572642ee468f15d708b7a9ad2452383f29e3368).

I think we can agree that it's strange that Mustache empty object is initialized with `global.Mustache`  (which is part of `this` context), but the `mustacheFactory` function is then passed a `Mustache` object (which is a part of `window` context).

Here is some sample code (the "mustacheJSCode" is a variable that would be replaced during the build time - with grunt or gulp for example):

```javascript
// Wrap Mustache.js into Foo namespace
var Foo = {};
(function() {
  <%= mustacheJSCode %>
}).apply(Foo);

// Foo.Mustache is now a valid object and window.Mustache is undefined as expected
```

simast:
Slightly updated the fix - should improve readability and will save a couple of bytes in minification.

dasilvacontin:
I think the latest change adds redundancy. If `Mustache` is meant to be initialized at `global.Mustache`, then we just pass `global.Mustache` to `factory`.

simast:
Reverted the fix to the first version.

dasilvacontin:
Okay, LGTM. Squash the commits and we'll merge. Thanks for the PR. :)

simast:
Done!

dasilvacontin:
Thanks!

PR #540, opened 2016-01-10
Add optional `output` argument to mustache bin

wizawu:
I am not sure whether this PR is welcome, since I didn't see much people had a requirement for using this argument.
Yet we have been using this feature in production (because `> output` is forbidden in certain cases).
Feel free to close it if you find it confusing.
Thanks.

### DIFF
--- a/test/_files/cli_output.txt
+++ b/test/_files/cli_output.txt
@@ -1,0 +1,1 @@
+Howdy LeBron, CLI rox
--- a/test/_files/cli_output.txt
+++ b/test/_files/cli_output.txt
@@ -1,1 +1,0 @@
-Howdy LeBron, CLI rox
--- a/bin/mustache
+++ b/bin/mustache
@@ -16,13 +16,14 @@ while ((partialArgIndex = process.argv.indexOf('-p')) > -1) {
 
 var viewArg = process.argv[2];
 var templateArg = process.argv[3];
+var outputArg = process.argv[4];
 
 if (hasVersionArg()) {
   return console.log(pkg.version);
 }
 
 if (!templateArg || !viewArg) {
-  console.error('Syntax: mustache <view> <template>');
+  console.error('Syntax: mustache <view> <template> [output]');
   process.exit(1);
 }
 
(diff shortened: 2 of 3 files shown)


phillipj:
Seems like a good feature 👍 Could you please add atleast one test for it aswell?

wizawu:
I think probably I shouldn't update the README. It might cause a mess :disappointed:

test/_files/cli_output.txt
```diff
@@ -0,0 +1 @@
+Howdy LeBron, CLI rox
```
phillipj: I'm thinking this file shouldn't be commited, it would even be better if it were deleted after the tests has run.

When this file exists in the repo, your changes in ./bin/mustache could actually be reverted, and the test below would still pass - right?

phillipj:
> Sorry, test added now.

Nice! One little comment about the output file, otherwise it LGTM.

wizawu:
Done :muscle:

phillipj:
Thanks! :+1:

dasilvacontin:
Btw @wizawu, what's the case in which `> output` is forbidden? I'm curious. :)

wizawu:
When using ansible's `shell` module.
```
- name: append
  shell: mustache .... >> output

- name: there won't be output
  shell: mustache .... > output
```
Apparently we don't want to append to output.

dasilvacontin:
@wizawu But does it raise an error, or?

Have you checked https://github.com/ansible/ansible/issues/9791 and/or http://docs.ansible.com/ansible/shell_module.html, specially:
```
# Run a command that uses non-posix shell-isms (in this example /bin/sh
# doesn't handle redirection and wildcards together but bash does)
- shell: cat < /tmp/*txt
  args:
    executable: /bin/bash
```

wizawu:
It won't raise an error. It would exit normally, but without any output. I know it is not a bug of mustache.

And yes, I've checked those docs. Thanks :) I just thought adding a third argument would be easier for me. I do agreed that it is just uncommon scene.

There is another case in package.json
```
{
  "scripts": {
    "build:html": "mustache view.json",
    "build:html:all": "ls html | xargs -I {} npm run build:html {} > dist/{}",
  }
}
```
The second script won't work as expected, since `> dist/{}` won't be **concated** to `build:html`.

I also agreed that it is another uncommon scene.

phillipj:
FYI if you'd like to send custom arguments to a npm script, like it seems you're trying to with `build:html`, you'll have to use double dash before any of the arguments arguments. E.g. `npm run build:html -- page.html`

Read more about it in [npm run script docs](https://docs.npmjs.com/cli/run-script#description).

wizawu:
@phillipj Very useful!

But after all, `>` won't be treated as an argument.

PR #551, opened 2016-02-27
Conditional Ternary at some points

anardy:
Changed to conditional ternary.

PR #552, opened 2016-03-09
Simplify multiple statements to a single line return statement

harryi3t:
I took [this PR](https://github.com/janl/mustache.js/pull/551)
and modified it to make it work.

Ran the test cases before pushing. All passed

### DIFF
--- a/mustache.js
+++ b/mustache.js
@@ -551,9 +551,7 @@
   };
 
   Writer.prototype.unescapedValue = function unescapedValue (token, context) {
-    var value = context.lookup(token[1]);
-    if (value != null)
-      return value;
+    return context.lookup(token[1]);
   };
 
   Writer.prototype.escapedValue = function escapedValue (token, context) {


harryi3t:
hey @phillipj,
could you look into this. Should I be pushing only one commit?

phillipj:
Thanks, I'll have a look as soon as I've got some spare time.

Please squash the commits, especially since the first commit introduces
syntax errors IIRC.

dasilvacontin:
Let me check if these changes provoke any edge case...

dasilvacontin:
I believe these changes introduces incorrect behaviour, but I'm waiting to confirm so: #553.

dasilvacontin:
I don't really think we can simplify these lines, unless we want to provide a different rendering behaviour for values different than `undefined` and `null`. And the result wouldn't be simpler or less lines of code. :smile:

PR #553, opened 2016-03-15
Assert null lookup when rendering an unescaped value

dasilvacontin:
It was not being asserted anywhere else, and I was suspecting that
PR #552 was changing current (and correct) behaviour.


dasilvacontin:
Gonna try placing the new assertion in a more adequate test file.

dasilvacontin:
Waiting for confirmation from @bobthecow, just in case: https://twitter.com/dasilvacontin/status/709796886568423424.

bobthecow:
This is probably in the category of things — like the definition of "falsey" — that is up to the individual library to decide what best fits in the host language. In my opinion the JavaScripty thing to do when encountering a `null` is to treat it as "the user intentionally put nothing here", and thus to interpolate it as an empty string. This is what Mustache.php does as well:

screen shot 2016-03-15 at 10 37 17 pm

Note that `null` in a block context will also stop context stack lookups from propagating.

At the very least, if I'm wrong it'd be a backwards compatibility break, so it's not allowed until the next major release :)

dasilvacontin:
> In my opinion the JavaScripty thing to do when encountering a `null` is to treat it as "the user intentionally put nothing here", and thus to interpolate it as an empty string.

> Note that `null` in a block context will also stop context stack lookups from propagating.

Thanks for your insight! We are doing both things – great! :smile:. We treat `null` as a marker for "nothing" (which is what `null` really is), so:
- the lookup function stops and returns `null` when it finds it (we fixed some edge-cases regarding this behaviour not long ago),
- it's considered as falsy for blocks, and
- it's currently being rendered as an empty string. :smile:

bobthecow:
:+1: sgtm

Issue #559, opened 2016-04-15
Tests failing on windows


kookookchoozeus:
I'm getting the following failed tests on windows (windows 8.1, node 4.4.0, if it matters). At a glance, it looks like piping file descriptors and not using backslashes for paths seem to cause most of these problems.

Relevant `npm test` output (excuse the formatting, you know how `cmd` is...):

```
  1) Mustache CLI writes syntax hints into stderr when runned with wrong number of arguments:
     Uncaught AssertionError: expected -1 to not equal -1
      at C:\Users\kookookchoozeus\mustache.js\test\cli-test.js:18:14
      at ChildProcess.exithandler (child_process.js:220:5)
      at maybeClose (internal/child_process.js:827:16)
      at Socket.<anonymous> (internal/child_process.js:319:11)
      at Pipe._onclose (net.js:475:12)

  2) Mustache CLI writes hints about JSON parsing errors when given invalid JSON:
     Uncaught AssertionError: expected -1 to not equal -1
      at C:\Users\kookookchoozeus\mustache.js\test\cli-test.js:25:14
      at ChildProcess.exithandler (child_process.js:220:5)
      at maybeClose (internal/child_process.js:827:16)
      at Process.ChildProcess._handle.onexit (internal/child_process.js:211:5)

  3) Mustache CLI writes module version into stdout when runned with --version:
     Uncaught AssertionError: expected -1 to not equal -1
      at C:\Users\kookookchoozeus\mustache.js\test\cli-test.js:32:14
      at ChildProcess.exithandler (child_process.js:220:5)
      at maybeClose (internal/child_process.js:827:16)
      at Process.ChildProcess._handle.onexit (internal/child_process.js:211:5)

  4) Mustache CLI writes module version into stdout when runned with -v:
     Uncaught AssertionError: expected -1 to not equal -1
      at C:\Users\kookookchoozeus\mustache.js\test\cli-test.js:39:14
      at ChildProcess.exithandler (child_process.js:220:5)
      at maybeClose (internal/child_process.js:827:16)
      at Process.ChildProcess._handle.onexit (internal/child_process.js:211:5)

  5) Mustache CLI without partials writes rendered template into stdout when successfull:
     Uncaught AssertionError: expected [Error: Command failed: C:\Windows\system32\cmd.exe /s /c "bin/mustache test/_fil
es/cli.json test/_files/cli.mustache"
'bin' is not recognized as an internal or external command,
operable program or batch file.
] to equal null
[...]

PR #558, opened 2016-04-14
Adds CLI support for partials with relative paths

kookookchoozeus:
Currently, if two partials with the same filename are passed through the CLI:

```bash```
mustache -p src/templates/header.mustache -p src/templates/common/header.mustache view.json template.mustache
```

only one of them is used (whichever's fs.createReadStream returns last) for the partial

```
{{> header }}
```

This change allows you to specify partials via paths relative to the template file

```
{{ ../common/header }}
```

PR #560, opened 2016-04-16
Fixes CLI tests for windows

kookookchoozeus:
Fixes https://github.com/janl/mustache.js/issues/559

All that was needed was a helper function for transforming *nix commands into windows commands:

* Makes sure `bin/mustache` is run with `node`
* Replaces `cat` with `type`
* Replaces forward slashes with backward slashes

Leaves commands as they are for *nix, so won't break the tests there

Also replaced the `Could not find file` strings to be compatible for windows

### DIFF
--- a/test/cli-test.js
+++ b/test/cli-test.js
@@ -2,20 +2,19 @@ require('./helper');
 
 var fs = require('fs');
 var path = require('path');
+var child_process = require('child_process');
 var _files = path.join(__dirname, '_files');
 var cliTxt = path.resolve(_files, 'cli.txt');
 var cliPartialsTxt = path.resolve(_files, 'cli_with_partials.txt');
 var moduleVersion = require('../package').version;
 
-var child_process = require('child_process');
-
 function changeForOS(command) {
 
   if(process.platform === 'win32') {
-    return command.
-      replace(/bin\/mustache/g, 'node bin\\mustache').
-      replace(/\bcat\b/g, 'type').
-      replace(/\//g, '\\');
+    return command
+      .replace(/bin\/mustache/g, 'node bin\\mustache')
+      .replace(/\bcat\b/g, 'type')
+      .replace(/\//g, '\\');
   }
 
   return command;
@@ -31,7 +30,6 @@ describe('Mustache CLI', function () {
   var expectedOutput;
 
   it('writes syntax hints into stderr when runned with wrong number of arguments', function(done) {
-
     exec('bin/mustache', function(err, stdout, stderr) {
       assert.notEqual(stderr.indexOf('Syntax'), -1);
       done();
@@ -101,17 +99,14 @@ describe('Mustache CLI', function () {
 
     it('writes it couldnt find template into stderr when second argument doesnt resolve to a file', function(done) {
       exec('bin/mustache test/_files/cli.json test/_files/non-existing-template.mustache', function(err, stdout, stderr) {
-        console.log(stderr);
-        assert.notEqual(stderr.indexOf('Could not find file:'), -1);
-        assert.notEqual(stderr.indexOf(changeForOS('test/_files/non-existing-template.mustache')), -1);
+        assert.isOk(/Could not find file: .+non-existing-template\.mustache/.test(stderr));
         done();
       });
     });
 
     it('writes it couldnt find view into stderr when first argument doesnt resolve to a file', function(done) {
       exec('bin/mustache test/_files/non-existing-view.json test/_files/cli.mustache', function(err, stdout, stderr) {
-        assert.notEqual(stderr.indexOf('Could not find file:'), -1);
-        assert.notEqual(stderr.indexOf(changeForOS('test/_files/non-existing-view.json')), -1);
+        assert.isOk(/Could not find file: .+non-existing-view\.json/.test(stderr));
         done();
       });
     });
--- a/test/cli-test.js
+++ b/test/cli-test.js
@@ -7,18 +7,23 @@ var cliTxt = path.resolve(_files, 'cli.txt');
 var cliPartialsTxt = path.resolve(_files, 'cli_with_partials.txt');
 var moduleVersion = require('../package').version;
 
-var exec = require('child_process').exec;
+var child_process = require('child_process');
 
 function changeForOS(command) {
 
   if(process.platform === 'win32') {
     return command.
       replace(/bin\/mustache/g, 'node bin\\mustache').
-      replace(/cat /g, 'type ').
+      replace(/\bcat\b/g, 'type').
       replace(/\//g, '\\');
   }
 
   return command;
+}
+
+function exec() {
+  arguments[0] = changeForOS(arguments[0]);
+  return child_process.exec.apply(child_process, arguments);
 }
 
 describe('Mustache CLI', function () {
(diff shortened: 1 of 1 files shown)


dasilvacontin:
Thanks @kookookchoozeus!

I think it would be cleaner to hijack `exec` calls so that you don't need to explicitly use `changeOS` around, alla:
```js
var fs = require('fs')

function exec () {
  arguments[0] = changeOS(arguments[0])
  return fs.exec.apply(fs, arguments)
}
```

test/cli-test.js
```diff
@@ -9,33 +9,46 @@ var moduleVersion = require('../package').version;
 
 var exec = require('child_process').exec;
 
+function changeForOS(command) {
+
+  if(process.platform === 'win32') {
+    return command.
+      replace(/bin\/mustache/g, 'node bin\\mustache').
+      replace(/cat /g, 'type ').
```
dasilvacontin: I think it would be preferable to wrap the regexes with `\b` symbols, like `/\bcat\b/`. Right now any word/filename ending with `cat` will instead end with `type`. It will likely not happen, but it's not as clean, I feel.

kookookchoozeus:
Thanks @dasilvacontin, very good suggestions that I should have realized myself, haha.

test/cli-test.js
```diff
@@ -83,14 +101,17 @@ describe('Mustache CLI', function () {
 
     it('writes it couldnt find template into stderr when second argument doesnt resolve to a file', function(done) {
       exec('bin/mustache test/_files/cli.json test/_files/non-existing-template.mustache', function(err, stdout, stderr) {
-        assert.notEqual(stderr.indexOf('Could not find file: test/_files/non-existing-template.mustache'), -1);
+        console.log(stderr);
+        assert.notEqual(stderr.indexOf('Could not find file:'), -1);
+        assert.notEqual(stderr.indexOf(changeForOS('test/_files/non-existing-template.mustache')), -1);
```
dasilvacontin: Could you explain why the assertion was split into two different ones?

test/cli-test.js
```diff
@@ -83,14 +101,17 @@ describe('Mustache CLI', function () {
 
     it('writes it couldnt find template into stderr when second argument doesnt resolve to a file', function(done) {
       exec('bin/mustache test/_files/cli.json test/_files/non-existing-template.mustache', function(err, stdout, stderr) {
-        assert.notEqual(stderr.indexOf('Could not find file: test/_files/non-existing-template.mustache'), -1);
+        console.log(stderr);
```
dasilvacontin: missed removing this `console.log`.

test/cli-test.js
```diff
@@ -83,14 +101,17 @@ describe('Mustache CLI', function () {
...
+        assert.notEqual(stderr.indexOf(changeForOS('test/_files/non-existing-template.mustache')), -1);
         done();
       });
     });
 
     it('writes it couldnt find view into stderr when first argument doesnt resolve to a file', function(done) {
       exec('bin/mustache test/_files/non-existing-view.json test/_files/cli.mustache', function(err, stdout, stderr) {
-        assert.notEqual(stderr.indexOf('Could not find file: test/_files/non-existing-view.json'), -1);
+        assert.notEqual(stderr.indexOf('Could not find file:'), -1);
+        assert.notEqual(stderr.indexOf(changeForOS('test/_files/non-existing-view.json')), -1);
```
dasilvacontin: Same here.

test/cli-test.js
```diff
@@ -7,13 +7,31 @@ var cliTxt = path.resolve(_files, 'cli.txt');
...
+  arguments[0] = changeForOS(arguments[0]);
+  return child_process.exec.apply(child_process, arguments);
+}
 
 describe('Mustache CLI', function () {
 
   var expectedOutput;
 
   it('writes syntax hints into stderr when runned with wrong number of arguments', function(done) {
+
```
dasilvacontin: You missed un-adding this line, right?

test/cli-test.js
kookookchoozeus: The filepath is absolute on Windows (e.g. C:\Users\you\path\to\mustache\test\_files...)

test/cli-test.js
```diff
@@ -7,13 +7,31 @@ var cliTxt = path.resolve(_files, 'cli.txt');
...
-var exec = require('child_process').exec;
+var child_process = require('child_process');
+
+function changeForOS(command) {
+
+  if(process.platform === 'win32') {
+    return command.
+      replace(/bin\/mustache/g, 'node bin\\mustache').
+      replace(/\bcat\b/g, 'type').
+      replace(/\//g, '\\');
```
dasilvacontin: When chaining, dots are usually put in the beginning of the line:
```js
if (...) {
  return command
  .replace(...)
  .replace(...)
  .replace(...)
}
```
Also, I'd say the blank lines inside the function are unnecessary. Specially the first one. Blank lines are mainly used to separate chunks / logic sections, and there's not much to separate here.

test/cli-test.js
```diff
@@ -7,13 +7,31 @@ var cliTxt = path.resolve(_files, 'cli.txt');
 var cliPartialsTxt = path.resolve(_files, 'cli_with_partials.txt');
 var moduleVersion = require('../package').version;
 
-var exec = require('child_process').exec;
+var child_process = require('child_process');
```
dasilvacontin: This import should go with the other Node core imports, under the `path` one. We group imports by:
- node core imports
- 3rd party libraries
- relative (same project) imports

dasilvacontin:
And I just noticed I used `fs` instead of `child_process` in my example. :')

test/cli-test.js
kookookchoozeus: Personally, I prefer having dots at the end of a line in these cases because 1) I'm paranoid about JS's semi-colon insertion (even though I've never had a semi-colon inserted where I didn't want one), and 2) just seeing `return command` sort of implies that that is all you want to return, having the dot at the end implies that's not the end of the statement.

That aside, I will, of course, make my changes consistent with the rest of the code base :)

test/cli-test.js
kookookchoozeus: Yeah, I was wondering why the `require('child_process').exec` was separate originally

test/cli-test.js
dasilvacontin: I see. That's not very obvious, just by reading the code..

I was going to suggest using a regex, eg `/Could not find file.*non-existing-template\.mustache/`., but don't worry about it, I'm going to refactor these assertions after we merge this PR (using `unexpected` or some nice assertion library).

kookookchoozeus: My initial thought was to use a regex as well, but then you'd also have to take the backslash problem into account with `[\/\\]{1,2}`

test/cli-test.js
dasilvacontin: Declaring a new var into which you make the modifications could solve both points:
```js
if (...) {
  var winCommand = command
    .replace(...)
    .replace(...)
    .replace(...)
  return winCommand
}
```
I've been writing JS without semicolons for over a year and I didn't have any problem, I swear! :)

test/cli-test.js
dasilvacontin: Not if you only add the last part of the file's path to the regex, you have no worries about the slashes that way. :)

test/cli-test.js
dasilvacontin: Oh, sorry, I didn't notice it was separate before. I didn't review the PR this file before it got merged, sadly.

kookookchoozeus:
Update: made the changes suggested above, hoping to get this merged soon so I can update https://github.com/janl/mustache.js/pull/558

dasilvacontin:
LGTM, thanks @kookookchoozeus! :)

PR #575, opened 2016-07-14
Document ways to set custom delimiters via JS

oughter:
I recently had the need to use custom delimiters and didn't want to do so within templates, which is the current documented approach.

### DIFF
--- a/README.md
+++ b/README.md
@@ -498,7 +498,7 @@ Custom delimiters can be used in place of `{{` and `}}` by setting the new value
 
 #### Setting in JavaScript
 
-The `tags` property of the `Mustache` object holds an array consisting of the opening and closing tag values. Set custom values by passing a new array of tags to `parse()`, which gets honored over the default values, or by overriding the `tags` property itself:
+The `Mustache.tags` property holds an array consisting of the opening and closing tag values. Set custom values by passing a new array of tags to `parse()`, which gets honored over the default values, or by overriding the `tags` property itself:
 
 ```js
 var customTags = [ '<%', '%>' ];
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -492,7 +492,31 @@ Mustache.render(template, view, {
...
 ```
 
-### Set Delimiter
+### Custom Delimiters
+
+Custom delimiters can be used in place of `{{` and `}}` by setting the new values in JavaScript or in templates.
+
+#### Setting in JavaScript
+
+The `tags` property of the `Mustache` object holds an array consisting of the opening and closing tag values. Set custom values by passing a new array of tags to `parse()`, which gets honored over the default values, or by overriding the `tags` property itself:
```
phillipj: > The `Mustache.tags` property holds an array ..

would be easier to grasp IMO

README.md
```diff
@@ -492,7 +492,31 @@ Mustache.render(template, view, {
...
+```js
+Mustache.parse(template, customTags);
+```
+
+##### Override Tags Property
+```js
+Mustache.tags = customTags;
+// Subsequent parse() and render() calls will use customTags
+```
+```
```
phillipj: This line could be deleted, as you've got the same on the previous line?

phillipj:
Thanks for the good contribution 👍 LGTM expect from some minor nits

README.md
```diff
@@ -492,7 +492,31 @@ Mustache.render(template, view, {
...
+
+#### Setting in Templates
```
bobthecow: something went wrong on this line (see the rendered output for this file)

README.md
bobthecow: it _needs_ to be deleted. it messes up the `Setting in Templates` heading :)

oughter:
Thanks for the feedback! Both items have been addressed.

phillipj:
Thanks 👍

PR #584, opened 2016-08-01
replace tokens == null with !tokens

Xuefeng-Zhu:


dasilvacontin:
Hi @Xuefeng-Zhu!

I personally prefer `(thing == null)`, since you can tell that it will be `true` only if `thing` is `null` or `undefined`, whereas with `(!thing)` you are opening up all the falsy values as possibilities (`null`, `undefined`, `0`, `''`, `NaN` ...). Ergo, you have to start looking into what kind of values `thing` can become, looking up what `cache` (in this case) can store, etc...

Not an improvement imo. Cheers!

PR #597, opened 2016-09-04
return mustache object from factory

mightyplow:
When using amdclean to create a footprint less javascript file, the object needs to be returned to let it properly be assigned to the created script variable.


dasilvacontin:
LGTM!

mightyplow:
Thank you very much!

dasilvacontin:
@mightyplow You're welcome :)! And sorry for the delay. >.<

PR #612, opened 2016-11-14
global.Mustache triggers JScript runtime error

nagaozen:
Because ASP/WSH script environments are language agnostic, one cannot assume it can be managed like a first-class javascript object. By changing the javascript global explicit assignment pattern to implicit assignment pattern (less is more) it works as expected.

Error before the change:
Microsoft JScript runtime error '800a01b6'
Object doesn't support this property or method
/lib/axe/classes/Parsers/mustache.asp, line 163

### DIFF
--- a/mustache.js
+++ b/mustache.js
@@ -11,8 +11,8 @@
   } else if (typeof define === 'function' && define.amd) {
     define(['exports'], factory); // AMD
   } else {
-    global.Mustache = {};
-    factory(global.Mustache); // script, wsh, asp
+    Mustache = {};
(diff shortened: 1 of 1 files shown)


phillipj:
Thanks for looking after us @nagaozen. I thought I remembered you got something similar merged last year -> https://github.com/janl/mustache.js/commit/60c2da80783a71409a8c3ecbc32a3d4e858df691.

Too ensure this regression doesn't sneek its way back in again, any thoughts about writing a unit tests for this fix?

Issue #562, opened 2016-04-19
Partial do not inherit indentation.


kuldeepdhaka:
```
var mustache = require("mustache")
var tmpl = "\t{{> test}}"
var data = {}
var partials = {"test": "Tested\nMulti line"}
console.log(mustache.render(tmpl, data, partials))
```

(NodeJS)

this printed

```
    Tested
Multi line
```

but i was expecting.

```
    Tested
    Multi line
```

kuldeepdhaka:
Beautiful code have proper indentation.

kuldeepdhaka:
https://github.com/mustache/spec/blob/master/specs/partials.yml#L13-L15

dasilvacontin:
Interesting, I didn't know about that part of the spec. We probably don't have any tests for that. Thanks for pointing it out!

brettstack:
Any update?

idefixcert:
could you fix this?
or ist there a workaround for that ?

balazser:
Could other mustache implementations solve it? 
This feature is necessary for generating indentation based blocks.

yotammadem:
I've added a proposed solution which fully complies with the partials spec.
The solution includes tests to all the scenarios described in the spec.
If anyone can review and merge this issue can be closed

yotammadem:
As indicated in the PR, I decided to take @kevindew 's solution and merged it with the partial spec tests I've added.  
Now it looks much better. 
Please review

phillipj:
This got fixed in https://github.com/janl/mustache.js/pull/705.

PR #613, opened 2016-12-03
Indent Partials

kevindew:
As referenced in this issue: https://github.com/janl/mustache.js/issues/562

Partials aren't indented as per the spec: https://github.com/mustache/spec/blob/master/specs/partials.yml#L13-L15

This adds indentation tracking for partials and applies them to the first instance of a partial on a line (to not indent inline partials which would violate a different part of the spec).

It's not really a perfect implementation - to really sort this would require reproaching how tokenising is done, so this sort of slots around the current approach to that.


brettstack:
Any update?

dasilvacontin:
Nope, all updates are in the respective Issues / PRs. We lack maintainer force. Any help reviewing issues / PRs is appreciated! :)

aayalam:
Is it possible to receive the indentation string within the called template?

I read the indentation of the first line is part of the specs, but It would be a reasonable solution to merely remove the indentation of the first partial's line and delegating this task to the called template; doing this way, the developer could decide to apply (or not) the indentation as needed.

phillipj:
These changes with some additions was merged in https://github.com/janl/mustache.js/pull/705, therefore closing this PR.

Thanks for your contribution 💯

PR #619, opened 2017-01-27
Pass context.view to partials function

taylorc93:
This idea arose from a project that I'm working on where I need to handle the logic for loading my own partials. This is because there might be hundreds of partials and I don't want to load all of them into memory (AKA using a partials object).

From parsing through the source code, I saw that the `partials` parameter for `renderPartial` can be a function. Great! However, when rendering through a function, I also need access to the context for the partial and currently, only the name of the partial is provided. This is where this PR comes is 😄.

### DIFF
--- a/mustache.js
+++ b/mustache.js
@@ -545,7 +545,9 @@
   Writer.prototype.renderPartial = function renderPartial (token, context, partials) {
     if (!partials) return;
 
-    var value = isFunction(partials) ? partials(token[1]) : partials[token[1]];
+    var value = isFunction(partials)
+      ? partials(token[1], context.view)
+      : partials[token[1]];
     if (value != null)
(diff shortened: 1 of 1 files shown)


phillipj:
Thanks for contributing 👍 

This sounds good to me. Could you add a test ensuring this works as intended?

taylorc93:
Sure thing! Busy tonight but will do this tomorrow

taylorc93:
Sorry for the delay, got caught up with other things. I noticed while trying to add a test that there was no check for retrieving partials using a function rather than an object, so that's what the test I wrote ensures. To do so however, I had to modify `test/render-test.js` slightly. Please advise if you have a better solution in mind or need anything else before merging.

PR #643, opened 2017-06-11
Writer.prototype.parse to cache by tags in addition to template string

### LINKED ISSUE #617: Writer.prototype.parse caches by template string, ignoring tags
I haven't run into actual issues with this, but I noticed that `Writer.prototype.parse` caches the result, where the cache key is just the template text. Therefore, these tests would fail:

```
describe('Mustache.parse', function () {
...

  describe('when parsing a template without tags specified then the same template with tags specified', function() {
    it('returns different tokens for the latter parse', function() {
      var template = "{{foo}}[bar]";
      var parsedWithBraces = Mustache.parse(template);
      var parsedWithBrackets = Mustache.parse(template, ['[', ']']);
      assert.notDeepEqual(parsedWithBrackets, parsedWithBraces);
    });
  });

  describe('when parsing a template with tags specified then a template with different tags specified', function() {
    it('returns different tokens for the latter parse', function() {
      var template = "(foo)[bar]";
      var parsedWithParens = Mustache.parse(template, ['(', ')']);
      var parsedWithBrackets = Mustache.parse(template, ['[', ']']);
      assert.notDeepEqual(parsedWithBrackets, parsedWithParens);
    });
  });

  describe('when parsing a template after already having parsed that template with a different Mustache.tags', function() {
    it('returns different tokens for the latter parse', function() {
      var template = "{{foo}}[bar]";
      var parsedWithBraces = Mustache.parse(template, ['(', ')']);

      var oldTags = Mustache.tags;
      Mustache.tags = ['[', ']'];
      var parsedWithBrackets = Mustache.parse(template);
      Mustache.tags = oldTags;

      assert.notDeepEqual(parsedWithBrackets, parsedWithBraces);
    });
  });
});
```

This issue is easily solved by changing `Writer.prototype.parse` to this (the change being to include the tags in the cache key):

```
  Writer.prototype.parse = function parse (template, tags) {
    var cache = this.cache;
    var tokens = cache[template];

    if (tokens == null)
      tokens = cache[template + ':' + (tags || mustache.tags).join(':')] = parseTemplate(template, tags);

    return tokens;
  };
```

1) It's unlikely that one would actually run into this bug in real life, though `parse` would be more correct if this simple fix was include. Do we want to fix this bug, even if it is unlikely to manifest?
2) One could argue that my suggested fix above breaks separation of concerns or some principle of abstraction. `Writer.prototype.parse` is simply a memoization of `parseTemplate` -- its only concern should be to cache the results of a parse. The actual details of how parsing is accomplished is in `parseTemplate`, and the fact that parsing falls back onto `mustache.tags` is one such detail. Referencing it in `Writer.prototype.parse` would be introducing a leak into the `parseTemplate` abstraction. Thoughts?

raymond-lam:
Resolve #617 
Because the results of `Writer.prototype.parse` depends on both the template string and the tags, it should cache by both

### DIFF
--- a/test/parse-test.js
+++ b/test/parse-test.js
@@ -103,4 +103,36 @@ describe('Mustache.parse', function () {
     });
   });
 
+  describe('when parsing a template without tags specified followed by the same template with tags specified', function() {
+    it('returns different tokens for the latter parse', function() {
+      var template = "{{foo}}[bar]";
+      var parsedWithBraces = Mustache.parse(template);
+      var parsedWithBrackets = Mustache.parse(template, ['[', ']']);
+      assert.notDeepEqual(parsedWithBrackets, parsedWithBraces);
+    });
+  });
+
+  describe('when parsing a template with tags specified followed by the same template with different tags specified', function() {
+    it('returns different tokens for the latter parse', function() {
+      var template = "(foo)[bar]";
+      var parsedWithParens = Mustache.parse(template, ['(', ')']);
+      var parsedWithBrackets = Mustache.parse(template, ['[', ']']);
+      assert.notDeepEqual(parsedWithBrackets, parsedWithParens);
+    });
+  });
+
+  describe('when parsing a template after already having parsed that template with a different Mustache.tags', function() {
+    it('returns different tokens for the latter parse', function() {
+      var template = "{{foo}}[bar]";
+      var parsedWithBraces = Mustache.parse(template, ['(', ')']);
+
+      var oldTags = Mustache.tags;
+      Mustache.tags = ['[', ']'];
+      var parsedWithBrackets = Mustache.parse(template);
+      Mustache.tags = oldTags;
+
+      assert.notDeepEqual(parsedWithBrackets, parsedWithBraces);
+    });
+  });
+
 });
--- a/test/parse-test.js
+++ b/test/parse-test.js
@@ -124,7 +124,7 @@ describe('Mustache.parse', function () {
   describe('when parsing a template after already having parsed that template with a different Mustache.tags', function() {
     it('returns different tokens for the latter parse', function() {
       var template = "{{foo}}[bar]";
-      var parsedWithBraces = Mustache.parse(template, ['(', ')']);
+      var parsedWithBraces = Mustache.parse(template);
 
       var oldTags = Mustache.tags;
       Mustache.tags = ['[', ']'];
--- a/mustache.js
+++ b/mustache.js
@@ -447,7 +447,7 @@
     var tokens = cache[template];
 
     if (tokens == null)
-      tokens = cache[template] = parseTemplate(template, tags);
+      tokens = cache[template + ':' + (tags || mustache.tags).join(':')] = parseTemplate(template, tags);
 
     return tokens;
   };


test/parse-test.js
```diff
@@ -103,4 +103,36 @@ describe('Mustache.parse', function () {
...
+      var parsedWithParens = Mustache.parse(template, ['(', ')']);
+      var parsedWithBrackets = Mustache.parse(template, ['[', ']']);
+      assert.notDeepEqual(parsedWithBrackets, parsedWithParens);
+    });
+  });
+
+  describe('when parsing a template after already having parsed that template with a different Mustache.tags', function() {
+    it('returns different tokens for the latter parse', function() {
+      var template = "{{foo}}[bar]";
+      var parsedWithBraces = Mustache.parse(template, ['(', ')']);
```
dasilvacontin: These are parenthesis, not braces, right?

raymond-lam: Yes, my bad. I'll fix it.

raymond-lam:
@dasilvacontin Addressed your comment -- good catch! I actually meant to test the template with `Mustache.tags` as the delimiters, with it being braces.

raymond-lam:
@dasilvacontin Awesome, thanks!

dasilvacontin:
I started refreshing my memory about the other PR, but I don't have enough time today. Will do tomorrow morning.

raymond-lam:
@dasilvacontin no rush! Thanks!

petrkoutnysw:
I am afraid this PR caused breaking change for us.
We are doing following:

1. parse our template with custom tags
2. render the template - according to mustache.js:465 our custom tags are not used, hence the tokens are different and custom tags are not recognized

phillipj:
@petrkoutnysw a bug fix landed after this got merged: https://github.com/janl/mustache.js/pull/664. This is also what's included in v2.3.1 of mustache.js.

Are you sure you're using the latest version? Or have I misunderstood and you have found another bug?

petrkoutnysw:
@phillipj Yeah I have 2.3.1. 

It is quite easy - if you take a look at parse function (line 465), it creates a cache using tags if cache does not exist already.  Then you call render function that calls parse internally without specifying tags. Thus it is getting wrong tokens. Hope it makes sense, I have not been reading mustache code previously.

Anyway, we froze our mustache to 2.3.0.

raymond-lam:
@petrkoutnysw Do you have replication steps (the template you used and how you made the call to render it)?

raymond-lam:
@petrkoutnysw 

It tried the following code:
```
mustache.parse('<<foo>>bar', ['<<','>>']);
mustache.render('<<foo>>bar', { foo:'biz'})
```

Which outputs "<<foo>>bar" as expected.

Then, if I do:
```
mustache.tags=['<<','>>']
mustache.render('<<foo>>bar', { foo:'biz'});
```

It outputs "bizbar", again as expected.

How are you setting the custom tags (there's only one way as far as I can tell) and calling the `render` function?

petrkoutnysw:
@raymond-lam Give me some time please, I am on conference and don't have time to reply right now. Thanks.

phillipj:
Maybe these two gives valuable input here? https://github.com/janl/mustache.js/issues/669 & https://github.com/janl/mustache.js/pull/662

raymond-lam:
@phillipj @petrkoutnysw Ah yes I see those reported issues now -- I'll have a look.

Issue #307, opened 2013-03-29
maybe strict escaping problem, forward slash turned into unicode entity


bryanrasmussen:
As per the subject, I am passing in uri, and the forward slashes come out as the entity #x2f;

IF it is strict escaping, is there a way to turn that off?

Issue #244, opened 2012-08-14
Unescape all values


ForbesLindesay:
would it be possible to add an option to not do any escaping of html at all?

NishadKumar:
> I haven't tested it, but according to the man page:
> `All variables are HTML escaped by default. If you want to return unescaped HTML, use the triple mustache: {{{name}}}.`
> If that doesn't work you could use a function which simply returns it's input unescaped.

Thanks a ton!! This works.

wll8:
Using `{{{` is null and void in this case:

``` js
template = `@{emoji}@{text}`
view = {
  text: "http://test.com",
  emoji: "✨",
}
mustache.render(template, view, {}, [`@{`, `}`])
```

lordbyron:
@wll8 perhaps what you're looking for, at least in the version i've seen, is `@{&text}`. the `&` symbol tells it not to do any escaping

PR #644, opened 2017-06-11
Document global Mustache.escape overriding capacity

paultopia:
Right now, the capacity to override HTML-escaping of strings globally isn't documented. Not only that, but the end of the main issue where this is discussed #244 proposes a solution that got implemented, but with a name that's slightly different from the actual user-facing name.  (The proposed solution is to overwrite `Mustache.escapeHTML`, but the actual name that client-facing code has to override is `Mustache.escape` --- which is only discovered if you dig down further and find #307. 

This capacity is super-useful, especially when combined with custom delimiters (where triple-mustache won't work). For example, I'm using Mustache to template LaTeX right now, and the capacity to override escaping globally is essential to keep me from going nuts. 

This PR just adds a line documenting this behavior.

### DIFF
--- a/README.md
+++ b/README.md
@@ -178,7 +178,9 @@ function loadUser() {
 
 The most basic tag type is a simple variable. A `{{name}}` tag renders the value of the `name` key in the current context. If there is no such key, nothing is rendered.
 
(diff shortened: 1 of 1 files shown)


jakob-bebop:
Thanks I needed just this :-)

sqren:
Bump. Time to merge? :)

phillipj:
For sure, thanks for the contribution and the friendly reminder 👍

Issue #659, opened 2018-05-04
Problem with custom delimiters


nkcjs:
|   Setting            |  Anticipant        |   Result    |
|------------------|------------------|------------|
| {{=<% %>==}} |      <%   %>=     |  <%  %> |

![image](https://user-images.githubusercontent.com/22137828/39612473-73095be8-4f92-11e8-9246-b023591c1036.png)

change to

```js
if (type === '=') {
  value = scanner.scanUntil(closingTagRe);
  if (value.charAt(value.length - 1) !== '=')
    throw new Error('Setting in template failure');
  value = value.slice(0, -1).trim();
}
```

scwong93:
I've submitted PR for this

phillipj:
As the support for custom delimiters got improved late 2018 in https://github.com/janl/mustache.js/pull/643 and subsequent PRs, this bug needs to be re-tested before considering a plausible fix.

phillipj:
Closing this for now due to lack of response from issue author. If it indeed still is an issue for what you're trying to do, don't hesitate re-opening the issue for further investigation.

PR #662, opened 2018-06-28
fix tag value if statement

scwong93:
Fixed the first if statement in getting tag value

### DIFF
--- a/mustache.js
+++ b/mustache.js
@@ -181,9 +181,10 @@
 
       // Get the tag value.
       if (type === '=') {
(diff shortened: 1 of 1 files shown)


mbrodala:
See #659

phillipj:
@scwong93 thanks!

Would be extremely valuable for us, right now and going forward, if we got a test or two in our test suite that reproduces a bug & verifies these changes fixes it.

phillipj:
Closing this for now as author has not responded to the request for tests.

PR #667, opened 2018-08-02
Fix browser test suite by using Node.js 6

phillipj:
For a long time we've struggled getting the browser test suite to run on [Sauce Labs](https://saucelabs.com/). Those tests were running on Node.js 4 which is getting quite old.

In a trial-n-error effort, I tried adding Node.js 6 as one of the versions to run on Travis CI, and at the same time ensure the browsers tests were run on that version instead of Node.js 4. Voila [the suite passed right away](https://travis-ci.org/janl/mustache.js/jobs/411477352), at least it did when in a feature branch, 🤞 it will still work when merged to master.

It therefore seemed like the issues we were seeing had to do with us running those tests via Node.js 4.

Also snook in a change for installing npm v2.x only when running tests on Node.js 0.8

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -6,7 +6,7 @@ node_js:
   - 4
   - 6
 before_install:
-  - npm install -g npm@2.x
+  - "test $TRAVIS_NODE_VERSION != '0.8' || npm install -g npm@2.x"
 script:
   - npm test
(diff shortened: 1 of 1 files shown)


phillipj:
Worth mentioning this is a precursor to releasing a couple of new mustache.js versions. I really want the test suite(s) to be running successfully on the master branch before cutting new versions.

PR #674, opened 2018-08-27
Changelog for v3.0.0

phillipj:
As we've got changes in master that we have agreed qualifies to a new major release, I put together a changelog draft for public scrutiny and would love feedback -- especially for migration guide included.

Any thoughts so far? Suggestions for a good corner case example for template cache and custom tags?

/cc @raymond-lam

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -8,7 +8,7 @@ This project adheres to [Semantic Versioning](http://semver.org/).
 We are very happy to announce a new major version of mustache.js. We want to be very careful not to break projects
 out in the wild, and adhering to [Semantic Versioning](http://semver.org/) we have therefore cut this new major version.
 
-The changes introduced will likely not require any actions for most using projects, the things to look out for that
+The changes introduced will likely not require any actions for most using projects. The things to look out for that
 might cause unexpected rendering results are described in the migration guide below.
 
 ### Major
@@ -25,7 +25,7 @@ might cause unexpected rendering results are described in the migration guide be
 
 #### Rendering properties of primitive types
 
-We have ensured properties of primitive types can be rendered at all times, that means `Array.length`, `String.length`
+We have ensured properties of primitive types can be rendered at all times. That means `Array.length`, `String.length`
 and similar. A corner case where this could cause unexpected output follows:
 
 View:
@@ -65,7 +65,7 @@ Output with v2.x:
 We have improved the templates cache to ensure custom delimiters are taken into consideration for the cache.
 This improvement might cause unexpected rendering behaviour for using projects actively using the custom delimiters functionality.
 
-Previously it was possible to use `Mustache.parse()` in a means to set global custom delimiters. If custom
+Previously it was possible to use `Mustache.parse()` as a means to set global custom delimiters. If custom
 delimiters were provided as an argument, it would affect all following calls to `Mustache.render()`.
 Consider the following:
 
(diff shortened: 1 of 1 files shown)


raymond-lam:
@phillipj Actually, I just checked out v2.3.2, and I got `There are 3 Stooges!`. I think the example needs object descent.

Here's an example from the test I wrote to address the original bug:
```javascript
Mustache.render('{{#singletonList}}{{singletonItem.length}}{{/singletonList}}', { singletonList: [{singletonItem: 'singleton item'}]});
```

v2.3.2 gives a blank string. `master` gives "14"

raymond-lam:
The original bug was stated as such:

`Mustache.render('{{#list}}{{a.length}}{{/list}}', {list: [{a: 'a'}]}) === ""` whereas `Mustache.render('{{a.length}}', {a: 'a'}) === "1"`

Originally the intent of the code was the first case: you can't access the length of a string. The original intent was for the second case also to give a blank string. It was "by luck" that you could access the length of a string. "Fixing the bug" would have meant taking the "feature" away, so we opted to "keep" the feature and fix the bug by making property access consistent.

raymond-lam:
@phillipj for the custom tag example, I think a good example might be any template using delimiters other than the default. The v2.3.2 read me suggests using parse as a means of specifying the delimiters for a template. I think it will be good to call out that this will no longer work, and that the guidance instead is simply to pass the custom tags into render directly.

raymond-lam:
@phillipj Not a waste of time at all. Happy to help!

phillipj:
I've fixed the bad primitive example and wrote the missing section about custom delimiters, PTAL.

CHANGELOG.md
```diff
@@ -3,6 +3,97 @@
 All notable changes to this project will be documented in this file.
 This project adheres to [Semantic Versioning](http://semver.org/).
 
+## [3.0.0] / x August 2018
+
+We are very happy to announce a new major version of mustache.js. We want to be very careful not to break projects
+out in the wild, and adhering to [Semantic Versioning](http://semver.org/) we have therefore cut this new major version.
+
+The changes introduced will likely not require any actions for most using projects, the things to look out for that
```
raymond-lam: Minor grammatical nitpick: comma splice; there should be either a period or semicolon between "projects" and "the things", rather than a comma.

CHANGELOG.md
```diff
@@ -3,6 +3,97 @@
...
+
+### Minor
+
+* [#673]: Add `tags` parameter to `Mustache.render()`, by [@raymond-lam].
+
+### Migrating from mustache.js v2.x to v3.x
+
+#### Rendering properties of primitive types
+
+We have ensured properties of primitive types can be rendered at all times, that means `Array.length`, `String.length`
```
raymond-lam: Similar grammatical nitpick: comma splice between "all times" and "that means" should instead be a period.

CHANGELOG.md
```diff
@@ -3,6 +3,97 @@
...
+  Larry:  characters
+  Curly:  characters
+```
+
+#### Caching for templates with custom delimiters
+
+We have improved the templates cache to ensure custom delimiters are taken into consideration for the cache.
+This improvement might cause unexpected rendering behaviour for using projects actively using the custom delimiters functionality.
+
+Previously it was possible to use `Mustache.parse()` in a means to set global custom delimiters. If custom
```
raymond-lam: Nitpick: probably should be "as a means" rather than "in a means".

CHANGELOG.md
```diff
@@ -3,6 +3,97 @@
...
+  })
+);
+
+>> TEST 1
+```
+
+The above illustrates the fact that `Mustache.parse()` made mustache.js cache the template without considering 
+the custom delimiters provided. This is no longer true.
+
+We don't longer encourage using `Mustache.parse()` for this purpose, but have rather added a fourth argument to
```
raymond-lam: Nitpick: "no longer encourage" rather than "don't longer encourage".

raymond-lam:
@phillipj Looks great! I left a few really minor grammatical nit-picky comments, but besides that I think this will serve as a good migration guide.

Thanks!

phillipj:
Thanks @raymond-lam, much appreciated feedback!

phillipj:
v3.0.0 is live on npm 🚀

PR #679, opened 2018-09-28
[#679]: Fix partials were not rendering tokens when not using default…

stackchain:
Issue found, when calling the parser inside partials it was not sending the tags to it, so the partials when not using the default tags {{, were not rendering properly


phillipj:
Awesome, thanks a lot for triaging this bug and fixing it right away! 👍

Landed in 864b5a5.

PR #705, opened 2019-07-09
Partials indentation

yotammadem:
A fix for: #562


phillipj:
Awesome, thanks for the quick response! I'm currently on holiday, will have
a look when I'm back.

phillipj:
Soooo I'm finally back from a long vacation + some general time off the interwebz, sorry for keeping you waiting so long, thanks for your patience.

mustache.js
```diff
@@ -592,12 +610,30 @@
...
 
     var value = isFunction(partials) ? partials(token[1]) : partials[token[1]];
-    if (value != null)
-      return this.renderTokens(this.parse(value, tags), context, partials, value);
+    if (value != null) {
+      var tagIndex = token[5];
+      var indentation = token[4];
+      var indentedValue = value;
+      if (tagIndex == 0 && indentation) {
+        var indentedValue = this.indentPartial(value, indentation);
```
phillipj: Only nitpick that got my brain worried at first; this kind of re-declaration of an already existing variable will explode when using `let | const`. Although that doesn't apply to `var` which is used here, I'm a fan of avoiding those kinds of questions from the get-go..

Bottom line removing `var` from this line would have avoided my initial worries and even make the intended purpose more clear (at least in my eyes).

Or did you very deliberately use `var` here for reasons I'm not aware of?

```suggestion
        indentedValue = this.indentPartial(value, indentation);
```

yotammadem: I agree, this was part of kevindew's commit which I didn't noticed. 
I am committing the suggestion

phillipj:
Great work and thank you so much for your patience! 👍 

FYI with the main intention of cleaning up the commits before landing, I got rid of the merge-with-master commit + latest `var` nitpick by rebasing with latest changes in master, before force pushing back into your branch.

yotammadem:
Thanks Phillipe for reviewing and merging the solution. Now I will finally be able to use the official mustsche module at my work place. (Until now I used a forked version)

phillipj:
v3.0.2 just got published to npm, so enjoy 🚀

Issue #627, opened 2017-04-10
Partials in nested objects


danielcobo:
It seems __partials don't work if they are stored in a nested object__.

I haven't seen docs detailing that partials are not working if they are stored in a nested object.
Is this a bug (in either mustache or my example) or is this expected behaviour?

__EXAMPLES:__

The following __works__ (see https://runkit.com/58db86c6eae8b80014630525/58eb71ee82d3930015030dad):
```node
var mustache = require("mustache")
    var view = {names : [
			{name : 'Foo'},
			{name : 'Bar'}
			]};
    var template = [
    		    '<h2>Names</h2>',
		    '{{#names}}',
		    	'{{>hello}}',
		    '{{/names}}',
		    ].join('');
    var partials = {
		    hello : '<strong>{{name}}</strong>'
		    }

mustache.render(template, view, partials);
```

while this __does not work__ (see https://runkit.com/58db86c6eae8b80014630525/58eb71ee82d3930015030dad): 
```node
var mustache = require("mustache")
    var view = {names : [
			{name : 'Foo'},
			{name : 'Bar'}
			]};
    var template = [
    		    '<h2>Names</h2>',
		    '{{#names}}',
		        '{{>hello.world}}',
		    '{{/names}}',
		    ].join('');
    var partials = {
		    hello : {
			    world : '<strong>{{name}}</strong>'
			    }
		    }

mustache.render(template, view, partials);
```

danielcobo:
I've **solved it by using a function to return partial template** instead of the object directly.
(See: https://runkit.com/npm/mustache)
```node
var mustache = require("mustache")
var objRef = require('objref');

var view = {names : [
                    {name : 'Foo'},
                    {name : 'Bar'}
                    ]};
var template = 	[
                '<h2>Names</h2>',
                '{{#names}}',
                    '{{>hello.world}}',
                '{{/names}}',
                ].join('');
var partialTemplates = {
                hello : {
                        world : '<strong>{{name}}</strong>'
                        }
                };

var partials = function(name){
    return objRef(partialTemplates, name, '.');
};

var result = mustache.render(template, view, partials);
```

This however is not the most elegant way. 
If the object of partials worked directly, it would be a much nicer solution.

ifgx:
Note: Your two first links to runkit are the same location.

danielcobo:
@ifgx sorry about that (the code examples here were correct though)

I've updated the link to reflect the code examples

[...]

PR #708, opened 2019-07-22
Expand context in renderPartial's call to renderTokens.

KnutRyagerInmeta:
Fix for "Partials in nested objects don't work #627"

This fixes the problem I mentioned in #627. Tests ran green.


KnutRyagerInmeta:
I have found an issue with call stack exceeding in another scenario of template expansion with this patch. So will look more into it. Maybe some veterans have more knowledge on how to safely expand the context?

Issue #712, opened 2019-08-26
Regression[3.0.2]: partials not working as intended after upgrading to version 3.0.2


wol-soft:
Hi,

after upgrading to mustache.js 3.0.2 a component of my application failed to render a template which uses partials.

I copied the affected parts of the application into a JSFiddle (https://jsfiddle.net/z1jq2pu5/13/) to reproduce the error. The current result of the fiddle looks like:
![image](https://user-images.githubusercontent.com/14750468/63687734-a8915800-c805-11e9-8e7a-01ee5ae157e6.png)

With version 3.0.1 it looked like (can be reproduced by simply changing the CDN URL which includes mustache.js into the JSFiddle):
![image](https://user-images.githubusercontent.com/14750468/63687770-c52d9000-c805-11e9-9931-9291d40a5bb4.png)

My application breaks with a message (which I can't reproduce in the JSFiddle somehow), may be helpful:
```
Error: Unclosed tag at 21
    at parseTemplate (VM448557 beerplop-game-1.59.3.min.js:17)
    at Writer.parse (VM448557 beerplop-game-1.59.3.min.js:17)
    at Writer.render (VM448557 beerplop-game-1.59.3.min.js:17)
    at subRender (VM448557 beerplop-game-1.59.3.min.js:17)
    at Object.<anonymous> (VM448557 beerplop-game-1.59.3.min.js:2810)
    at Writer.renderSection (VM448557 beerplop-game-1.59.3.min.js:17)
    at Writer.renderTokens (VM448557 beerplop-game-1.59.3.min.js:17)
    at Writer.renderPartial (VM448557 beerplop-game-1.59.3.min.js:17)
    at Writer.renderTokens (VM448557 beerplop-game-1.59.3.min.js:17)
    at Writer.render (VM448557 beerplop-game-1.59.3.min.js:17)
```

Seems like the tags of the partial aren't parsed correctly

phillipj:
Thanks a lot for reporting! 👍At first glance, this sounds like something we should be able to reproduce in a unit test, both for the sake of fixing it ASAP and to ensure it doesn't happen again later.

wol-soft:
Hi @phillipj, the values passed to functions are shifted by the indentation of the partial, see the tests from #713

phillipj:
For future reference, this regression got introduced by https://github.com/janl/mustache.js/pull/705

phillipj:
v3.0.3 just got published to npm with a fix for this regression.

[...]

PR #713, opened 2019-08-26
Add test cases for custom functions in partials

wol-soft:
Adds test cases to reproduce the issue https://github.com/janl/mustache.js/issues/712


phillipj:
Perfect, much appreciated! 💯

Issue #715, opened 2019-08-27
An inline partial should not be indented


hail2u:
The content of an inline partial is indented in v3.0.2 (of course not in v3.0.1). People does not want these extra white spaces in an HTML file.

test.json:

```json
{}
```

test.mustache:

```mustache
    <div>{{> partial}}</div>
```

partial.mustache:

```mustache
This is a partial.
```

Run:

```bash
$ mustache -p partial.mustache test.json test.mustache
    <div>    This is a partial.</div>
```

phillipj:
Thanks for reporting!

I for sure agree your example looks both weird and unexpected. Without digging deep into how this could be solved in this project, I'm starting to get curious how other mustache implementations (java, PHP etc) handles indentation of partials.

The expected (or shall we say sensible) output of your example would be great to have as an unit test inside [./test/partial-test.js](https://github.com/janl/mustache.js/blob/master/test/partial-test.js).

phillipj:
@yotammadem you got any thoughts on this by any chance?

yotammadem:
I will take a look at this soon. Strange that the partials spec does not test it

yotammadem:
This is strange..
When I tried to reproduce this with a unit test:
```
    it('Inline partials should not be indented', function () {
      var template = '"    <div>{{> partial}}</div>"';
      var data = {};
      var partials = {'partial':'This is a partial.'};
      var expected = '"    <div>This is a partial.</div>"';
      var renderResult = Mustache.render(template, data, partials);
      assert.equal(renderResult, expected);
    });

```
The unit test passed..
I will try with the command line

yotammadem:
Ok, it was my mistake, when I ran this test, it failed:
```
it('Inline partials should not be indented', function () {
      var template = '    <div>{{> partial}}</div>';
      var data = {};
      var partials = {'partial':'This is a partial.'};
      var expected = '    <div>This is a partial.</div>';
      var renderResult = Mustache.render(template, data, partials);
      assert.equal(renderResult, expected);
    });
```

yotammadem:
@phillipj  Take a look at: https://github.com/janl/mustache.js/pull/716

PR #716, opened 2019-08-27
Fix indentation of inline partials (issue #715)

yotammadem:


phillipj:
Cool, I'll have a look as soon as I -- hopefully before the weekend.

phillipj:
Interesting solution indeed!

Thinking out loud, my first hope was that we could fix this in a very isolated part of the source code, instead of involving several functions and new kinds of data/tokens being created. E.g. only in the `Writer.prototype.indentPartial()` function would have been ideal, as that function sounds very much in charge for indentation of partials.. Might be easier said than done though.

Thanks a lot for your quick response. I'll dive in as well to see if we end up with the same solution or maybe a combination of both.

yotammadem:
I agree. I tired to find the quickest solution that satisfies our requirements. It emphasises the complexity of having too much coupling between the parse function and the indentPartials function like you said. I will try to figure out a better way, now that we have all tests in place, it should be easy to verify the solution.

phillipj:
Thanks a lot for finding a fix to this indentation bug! 👍

phillipj:
Published in v3.1.0.

Issue #698, opened 2019-05-29
ES Module support


zekth:
Hello!

I was wondering if there is a plan to support an ES module version of **Mustache.js**? If so i can contribute to it.

**Why?**
I'm working on the [Deno](https://github.com/denoland/deno) project, and there is a discussion about templating (Deno only supports ES modules). Problem is there is no template renderer at the moment. One idea was to port **mustache** to **Deno**, so a port means a different code stream which needs maintainance. Having it directly in mustache would be a benefit for both projects i think. But in the case of mustache it means having a `mustache.mjs` added to the repo,**edit**: or change the mustache.js to make it ES module compatible.

Your thoughts?

ref: https://github.com/denoland/deno_std/issues/391

phillipj:
Hi @zekth!

Exciting plans you've got with deno going forward. I'm very positive to making this project ES module compatible, and completely agree with your thoughts on avoiding a port.

You got any concrete thoughts what would be necessary to make this happen?

zekth:
Here is a (really ugly) port i've done this morning (in typescript) https://github.com/zekth/deno_mustache/blob/master/mod.ts

i've ported some test just for the beginning.
The only concern i have is all the different environments you support with mustache, having the main js file ES module compatible is better than another `mjs` i think but i don't want to break anything :)

phillipj:
Thanks a lot for the reference! 👍 

Since I've got zero experience with deno, I'll fire off some trivial questions to trigger some discussions:

1. What requirements to deno have to make this work? Does it need to be a `.mjs` file or does it care about the [`type="module"` of `package.json`](https://nodejs.org/api/esm.html#esm_code_package_json_code_code_type_code_field)?
2. Any file naming requirements?
3. Does TypeScript matter here somehow?

..and so on, any for-dummies-clues would be much appreciated. At the moment I've got too little knowledge about deno to think creatively about plausible ways of tackling this.

zekth:
1. There is no special requirement as it only needs the library to be an ES module. Example `lodash` works without any porting. Loading of the module is made throught HTTP fetching, no package json or anything.
2. No naming convention
3. You can use TypeScript of JavaScript, Deno handles both.

For example using mustache in deno would look like:
```ts
import * as mustache from 'https://raw.githubusercontent.com/janl/mustache.js/master/mustache.js'
mustache.render(......
```

If you want more infos you can check this talk from Ryan: https://www.youtube.com/watch?v=z6JRlx5NC9E

[...]

PR #728, opened 2019-11-06
Transition source code to be an ES module

phillipj:
**TDLR**; allow Node.js and [Deno](https://https://deno.land) to use this package as a native [ECMAScript module](https://nodejs.org/dist/latest/docs/api/esm.html).

Up until now, we've very deliberately held back on introducing any kind of build step to this project. Personally think that's been a wise decision as the benefits up doing so has not been obvious in terms of the complexity such a build step usually introduces in a project. Keeping everything in one .js file in good old fashioned ES3 has worked out really well so far.

That changes with what's introduced in this PR.

Why on earth should we suddenly change what has worked well all these years?? To me it's about following along with how both Node.js & JavaScript evolves and reacting to new related runtimes emerging, like Deno is a great example of.

Node.js, Deno and [several modern browsers](https://caniuse.com/#feat=es6-module) has built-in ES module support. Hence it's only natural to consider following along and make this package an ES module, rather than holding back just for the sake of not introducing a build step for the purity it. 

Ofcourse backwards compatibility with non-ES module systems are kept as is, this is where a build step comes into play. The source code is now put inside the `mustache.mjs` file, and `mustache.js | mustache.min.js` is the build step output.

Tried my best creating a sensible commit history with some in-depth reasoning for each of them. Hopefully they'll shed some more light on the rationale and be of some help going forward (if we decide to merge this).

<hr>

Missing pieces being worked on:

- ~[ ] `type: module` in `package.json` as described in [Node.js ECMAScript Modules docs](https://nodejs.org/dist/latest/docs/api/esm.html#esm_code_package_json_code_code_type_code_field)~
- [x] make CI module-systems test for Node.js using it as an ES module

**UPDATE:**

After reading the details about Node.js' behaviour when `type: module` is set, that does not seem appropriate for us. It would result in making Node.js interpret `mustache.js` as an ES module because it's set as the `main` file. That in combination with us wanting to keep that file backwards compatible with CommonJS & UMD, `mustache.js` will have to stay non-ES Module.

Fixes https://github.com/janl/mustache.js/issues/698


phillipj:
@zekth & friends; a few weeks has passed since I ran CI, and after re-running the test that verifies mustache.js works with deno now, it exploded with what seems like TypeScript errors to me;

```
$ deno test --allow-net=deno.land test/module-systems/deno-test.ts

...

Compile file:///home/runner/work/mustache.js/mustache.js/test/module-systems/deno-test.ts
Download https://deno.land/std/testing/mod.ts
Download https://deno.land/std/testing/asserts.ts
Download https://deno.land/std/fmt/colors.ts
Download https://deno.land/std/testing/diff.ts
Download https://deno.land/std/testing/format.ts
error TS2371: A parameter initializer is only allowed in a function or constructor implementation.

► https://deno.land/std/testing/asserts.ts:132:39

132 export function assert(expr: unknown, msg = ""): asserts expr {
                                          ~~~~~~~~

error TS2304: Cannot find name 'asserts'.

...
```

https://github.com/janl/mustache.js/pull/728/checks?check_run_id=316803283#step:5:44

Any thoughts what might be causing this? Would be great to know we don't merge an unstable test that might fail if it hasn't run for a while..

I wouldn't be surprised at all if it's because I'm doing something wrong, either in the [related `deno` CI setup](https://github.com/janl/mustache.js/pull/728/files#diff-035ccb988bcea79c1f0f26da5409a727R75) or executing the test via `deno`.

test/module-systems/deno-test.ts
```diff
@@ -0,0 +1,17 @@
+import { test } from "https://deno.land/std/testing/mod.ts";
+import { assertEquals } from "https://deno.land/std/testing/asserts.ts";
```
zekth: You're not aiming the version in here.
CI is running using `0.21.0` let's try using `/std@v0.21.0/` here?

phillipj:
Ahh! I'm assuming what I'm doing is not best practice in terms of importing
without a hint of version number anywhere. Should it be part og the import
or some external file, like package.json for npm projects?

zekth:
By this way you import the Latest version which is not recommended because your runtime and your imports may have incompatibility. Like using node 8 with node std lib of node 12

phillipj:
Yupp, that struck me too with your first answer.

So currently this is always controlled by specifying the version in each
and every import in a project, or is there some kind module management
alternative as well that I should be aware of?

zekth:
Yes you can use import maps: https://github.com/denoland/deno/issues/3268#issuecomment-551859728

phillipj:
Thanks, much appreciate your quick replies on this! I'll have a look at
improving it later tonight.

test/module-systems/deno-test.ts
phillipj: `/std@v0.21.0/` worked like a charm, thx!

phillipj:
🛑 found after having a Node.js test in place, with v13.2.0 that has .mjs support enabled by default: [esm-usage test result](https://github.com/janl/mustache.js/pull/728/checks?check_run_id=317640238).

That CI test result points to a big difference compared to how the module resolution in Node.js has worked until now; .json files cannot be `import`ed directly. This is outlined in the [Node.js / ESM / Experimental JSON Modules](https://nodejs.org/dist/latest-v13.x/docs/api/esm.html#esm_experimental_json_modules) docs section, but doesn't help much as of writing this, because that is experimental and therefore opt-in with CLI flags. Personally would really appreciate if we wouldn't end up having to tell Node.js ES Modules users to remember to enable experimental features.

Therefore also worth consider other ways to getting the current version number into the source code. Or even ask ourselfs if having that `.version` property is really valuable to anyone.

Personally feel we need to sort this out before merging to master & publishing anything to npm.

zekth:
> Personally would really appreciate if we wouldn't end up having to tell Node.js ES Modules users to remember to enable experimental features.

Definitely agree.

Is injecting the version in the build process thinkable ? (it's a bit ugly compared to importing from package.json)

phillipj:
Also considering that even if Node.js ES modules would have .json support, I noticed the docs or some GitHub issue I stumbled upon, said that named imports would not be supported. In practise that would mean that we would have to `import` the entire contents of `package.json`, which for the built output is not ideal, when the only thing we really want is the rather simple `version` field value.

I'm not too familiar with rollup.js to be honest, might be a very natural way of doing that as part of the build process as you suggest 🤷‍♂

zekth:
> I'm not too familiar with rollup.js to be honest, might be a very natural way of doing that as part of the build process as you suggest 🤷‍♂

We can use this in the build process: https://github.com/rollup/plugins/tree/master/packages/replace

phillipj:
Nice!

That definitively looks like what we need to make sure old the `.version` field is kept for non-ES module systems. Although won't have much effect for the ones using the `.mjs` version 🤔 Which makes question the added value of that field to be honest.

Maybe there's some traces of the reason it was introduced in the first place if we dig into the git history / GitHub PRs.

phillipj:
On second thought, we've managed to survive all these years with the `.version` field value being hard coded in the source file (`mustache.js`). It has been kept in sync by a `git pre-commit hook`: https://github.com/janl/mustache.js/blob/e52478c14bcc035c09595cedffed5b2b870d8e17/hooks/pre-commit#L27-L32

Maybe we should keep that ad-hoc tool for now. ESm-ifying isn't really about refactor that `.version` anyways, so I realise I might have jumped that gun prematurely.

phillipj:
With all known issues fixed, commits cleaned up and PR no longer marked as a draft, this is very close to getting merged.

No objections has been made and we've got a handful of tests in place verifying usage from module systems we want to officially support.

phillipj:
For the sake of testing the pre-commit in practise, v3.2.0-beta.0 has been published to npm. Tweaks had to be made to that pre-commit, so good to have that checked off before merging as well.

Will be verifying the contents of that beta version as the last frontier before pressing the o-mighty-merge button.

Sorry for being slow, I just want to be as certain as possible this won't cause havoc knowing this is a big change in terms of how this module is glued together. Quite surprised we have 1.3 million downloads a week from npm these days, that deserves respect when pushing new versions IMO 😃

phillipj:
Publish to npm as v3.2.0, with [updated CHANGELOG.md](https://github.com/janl/mustache.js/blob/master/CHANGELOG.md#320--18-december-2019).

@zekth thanks for pushing this forward and being incredibly patience! Hoping the deno community will get some value of this going forward 🤞

Issue #730, opened 2019-12-03
Allow internal cache to be disabled


AndrewLeedham:
I am running in to memory limits when compiling lots of mustache.js templates in one go (for testing purposes). I am wanting to disable mustache's internal cache and implement my own that would remove stale templates. This is currently not possible, the only API seems to be cache clearing. Is it something that has been considered for this project? If not does it seem like it fits with the library? Happy to put in a PR.

Possible workarounds:
* Clear the cache after every component render, this seems inefficient.
* Monkey patch the `Writer.prototype.parse` function to skip the caching. This is the best current solution but having it as a first class API would be better.

phillipj:
That's an interesting question!

I remember there has been questions about making the cache smarter and offer different cache strategies, but we've been reluctant to implement & maintain that as part of this package.

Opening up for allowing others to override the caching strategy is a different story, and probably something we should have asked ourselfs before? 🤷‍♂ Totally agree to your suggestion of exposing this first class, rather than forcing others to monkey patch the inner workings.

I'd happily review a PR that enables this 👍 In general, as few & focused changes as possible would be better, compared to turning the source upside down to make this happen. Let me know if you have more questions or want input.

AndrewLeedham:
Thanks for getting back to me @phillipj. I have put a PR in with a basic disabling mechanism, but it does not particularly address the monkey patching concern, just makes it easier. See https://github.com/janl/mustache.js/pull/731/commits/88457abb531a2c113dd5daf2cb39597c51f14ea4#diff-66c3acc3931379ca3437082f30ec7aecR175-R185 for an example of how a user would override the cache.

Perhaps a better approach would be to split the cache retrieval into a separate function, then expose a function that accepts a predicate which handles caching?

phillipj:
Your PR surely looks like it deliveries on the need for _disabling the cache_ completely when needed.

> Perhaps a better approach would be to split the cache retrieval into a separate function, then expose a function that accepts a predicate which handles caching?

That's also an interesting idea! Thinking out loud without considering how that would be implemented practise, extracting the cache retrieval & storing into separate methods could potentially allow for other caching strategies to be provided, without those strategies having to be implemented in the mustache.js itself.

Building upon that idea and wanting to disable the cache, that could be accomplished by providing either a predicate as you mention or a no-op cache storing method 🤔 

Whether we actually **need** and should think about solving these use cases is up for discussion though.

AndrewLeedham:
[...]

PR #731, opened 2019-12-10
Allow template caching to be customised

AndrewLeedham:
fixes: #730 

Not 100% sold on this approach but this allows for `parse` to be more cleanly overridden without having to clear the cache for every call. Open to suggestions.


phillipj:
Cool @AndrewLeedham! Thanks for submitting an approach to disabling the cache when needed 👍

For the sake of trying to keep the related discussion in one thread, I posted a longer reply in https://github.com/janl/mustache.js/issues/730#issuecomment-566419314.

AndrewLeedham:
@phillipj I have implemented your recommendations from #730, took a bit to get the tests to pass, but all is working now.

Should this stay undocumented?

test/parse-test.js
```diff
@@ -58,6 +58,18 @@ var expectations = {
...
+    set: function set (key, value) {
+      this._cache[key] = value;
+    },
+    get: function get (key) {
+      return this._cache[key];
+    },
+    clear: function clear () {
+      this._cache = {};
+    }
+  };
```
phillipj: Hmz, I'm a bit concerned this will prevent us from testing the default built-in cache found in the source (`mustache.mjs`). Cause if we do some adjustments to that default cache sometime in the future, those adjustments will be overwritten by this custom template cache, right?

Guessing this trick is to reset the template cache back to (what's currently) the default cache, especially important after the custom template cache test has run, right?

If so, maybe it would be possible to:

1. get a hold of the *actual default* template cache in a `before()`
2. store a reference to that original template cache object in a top-level variable
3. stick that reference back as part of this `beforeEach()`, something like `Mustache.templateCache = originalTemplateCache`

..or have I totally misunderstood the purpose of this?

mustache.mjs
```diff
@@ -479,14 +479,25 @@ Context.prototype.lookup = function lookup (name) {
...
+    }
+  };
 }
 
 /**
  * Clears all cached templates in this writer.
  */
 Writer.prototype.clearCache = function clearCache () {
-  this.cache = {};
+  this.templateCache !== undefined && this.templateCache.clear();
```
phillipj: Let's do this undefined-check the same way for consistency, meaning here and on line 511. Up to you which approach you prefer 😃

phillipj:
Great job ensuring these changes works on legacy aswell 👍 

> Should this stay undocumented?

Good question! As setting a custom template cache is only needed for a selected few, I'm leaning towards not documenting disabling / custom cache in the `README` until we get lots of requests about how it works...

Although *some* explanation would probably be valuable, both for us selfs and future collabs, maybe it would be fitting to explain the concept as a block comment above the `set templateCache` that gets introduced here?

mustache.mjs
AndrewLeedham: Done, I had a quick look at the eslint docs to see if there was a rule to enforce this. But the closest is `no-undef` which also stops `undefined` from being used to initialise objects etc. So I have raised a new rule proposal in the ESLint repo: https://github.com/eslint/eslint/issues/12754

AndrewLeedham: Done, I had a quick look at the eslint docs to see if there was a rule to enforce this. But the closest is `no-undef` which also stops `undefined` from being used to initialise objects etc. So I have raised a new rule proposal in the ESLint repo: https://github.com/eslint/eslint/issues/12754

test/parse-test.js
AndrewLeedham: Done, I had a quick look at the eslint docs to see if there was a rule to enforce this. But the closest is `no-undef` which also stops `undefined` from being used to initialise objects etc. So I have raised a new rule proposal in the ESLint repo: https://github.com/eslint/eslint/issues/12754

AndrewLeedham: You understood correctly, good catch. I have updated this, had to provide a getter for the Writer `templateCache` as well to facilitate this.

phillipj:
These changes has been published to npm as [v4.0.0](https://github.com/janl/mustache.js/blob/master/CHANGELOG.md#400--16-january-2020).

Thanks a lot for the great initiative and this PR 👍

PR #733, opened 2019-12-30
Allow the CLI to use JavaScript views when the project has ES6 modules enabled.

### LINKED ISSUE #732: Cannot use JavaScript view from CLI when project has  "type": "module"
If the package.json for a project has enabled ES6 modules by including
```json
  "type": "module",
```
then it is not possible to use a JavaScript view.

If you try to create a view with a `.js` extension then you get an error such as the following:
```
(node:14179) Warning: require() of ES modules is not supported.
require() of /home/eobrain/src/mergi/template/index_view.js from /home/eobrain/src/mergi/node_modules/mustache/bin/mustache is an ES module file as it is a .js file whose nearest parent package.json contains "type": "module" which defines all .js files in that package scope as ES modules.
Instead rename index_view.js to end in .cjs, change the requiring code to use import(), or remove "type": "module" from /home/eobrain/src/mergi/package.json.
internal/modules/cjs/loader.js:1163
      throw new ERR_REQUIRE_ESM(filename);
```
If you try to solve the problem by renaming the view to have a `.cjs` suffix you get an error such as the following
```
Shooot, could not parse view as JSON.
Tips: functions are not valid JSON and keys / values must be surround with double quotes.

SyntaxError: Unexpected token m in JSON at position 0
```
because it is not recognized as being JavaScript and it is being parsed as JSON.

eobrain:
Fix #732 by allowing CLI JavaScript views to have the `.cjs` suffix.

### DIFF
--- a/bin/mustache
+++ b/bin/mustache
@@ -131,7 +131,8 @@ function isStdin (view) {
 }
 
 function isJsFile (view) {
-  return path.extname(view) === '.js';
+  var extension = path.extname(view);
+  return extension === '.js' || extension === '.cjs';
 }
 
 function wasNotFound (err) {
--- a/test/_files/cli.cjs
+++ b/test/_files/cli.cjs
@@ -1,0 +1,3 @@
+module.exports = {
+  name: 'LeBron'
+};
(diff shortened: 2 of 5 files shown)


phillipj:
Thanks a lot for getting `.cjs` supported both in the CLI and in the test suite in general.

v3.2.1 just got published to npm 🚀 Happy new years!

Issue #738, opened 2020-02-24
Custom delimiters do not work in nested partials


aielo:
Custom delimiters only work in the template itself and its (level 1) partials (thanks to #679). For inner partials (level 2+), however, default tags are used rather than the custom ones provided in `render`.

This behavior can be reproduced with the **sample code**:
```
const Mustache = require("mustache");
const tags = ["[[", "]]"];
const partials = {};

for (let l = 1; l <= 5; l++) {
  partials["level" + l] = `partial level ${l}\n`;
  if (l < 5) {
    partials["level" + l] += `${tags[0]}> level${l + 1} ${tags[1]}`;
  }
}

const output = Mustache.render(
  `template\n${tags[0]}> level1 ${tags[1]}`,
  {}, partials, tags
);

console.log(output);
```
**Double brackets** are set as custom tags in `const tags = ["[[", "]]"]`, and it outputs:
```
template       
partial level 1
partial level 2
[[> level3 ]]
```
`[[> level1 ]]` in template and `[[> level2 ]]` in the first level partial are properly resolved, while the first inner partial `[[> level3 ]]` is left unresolved.

Switching to `const tags = ["[[{{", "}}]]"]`, and keeping the **curly brackets within custom delimiters**, it outputs:
```
template
partial level 1      
partial level 2      
[[partial level 3    
  [[partial level 4  
    [[partial level 5
]]]]]]
```
Here, custom tags are again used in template and first level partial. Default Mustache tags are then used in inner partials, rendering them wrongly.

**Overriding `Mustache.tags`** works normally. Changing the sample code:
```
...
Mustache.tags = ["[[", "]]"];
const tags = Mustache.tags; // just to generate test partials
...
const output = Mustache.render(
  `template\n${tags[0]}> level1 ${tags[1]}`,
  {}, partials // no custom tags
);
...
```
Output is the expected:
```
template       
partial level 1
partial level 2
partial level 3
partial level 4
partial level 5
```

phillipj:
Very sorry for the late response, been more busy than usual lately.

Thanks a bunch for the very elaborate description of what you've stumbled onto and steps to reproduce, much appreciated 👍

PR #739, opened 2020-02-25
Fix custom delimiters in nested partials

aielo:
Fix to #738.

Changes:

- Parameter `tags` added to `renderTokens` calls from `renderPartial`
- `Nested partials should support custom delimiters.` test added to `partial-test.js`


phillipj:
v4.0.1 published 🚀

PR #749, opened 2020-04-07
docs: Fix simple typo, skiped -> skipped

timgates42:
There is a small typo in test/mustache-spec-test.js.

Should read `skipped` rather than `skiped`.


phillipj:
Good catch, thanks! 👍

PR #755, opened 2020-05-31
Update and simplify package.json

friederbluemle:
Add `.idea/` to `.gitignore`, and a couple of minor updates and simplifications to `package.json`:

- npm _always_ includes `LICENSE`, and the file declared in `main` in the packaged artifact, therefore, it's not necessary to also list them in `files` (reference: https://docs.npmjs.com/files/package.json#files)
- Also removed the engines section which listed a suuuper old version of npm.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -7,5 +7,5 @@ qooxdoo.mustache.js
 test/render-test-browser.js
 npm-debug.log
 .DS_Store
-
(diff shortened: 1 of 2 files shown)


phillipj:
Thanks a lot for doing some much needed chores & cleanup of `package.json`, appreciated 👍

PR #777, opened 2021-04-09
Fix missing semicolon in `test/cli-test.js`

Mejia1994:
when run `npm test` it shows 

> test\cli-test.js 13:85  error  Missing semicolon  semi


phillipj:
Thanks for pointing this out!

It should obviously have been caught in the CI runs, but you made me aware all `test/*.js` files was lacking in our linting step.

Fixed that now (in https://github.com/janl/mustache.js/commit/af216b0ce7ec6e4e384a2c43f0f65e2a31900ca0) so this won't happen again.

PR #812, opened 2023-01-18
Makes README.md examples more convenient

tahashieenavaz:


phillipj:
Thx for making the examples more inline with modern practices @tahashieenavaz 👍🏻

PR #828, opened 2024-05-13
Dynamic names

lucas42:
The mustache spec now includes an optional part of the spec for using dynamic names in partials: https://github.com/mustache/spec/blob/master/specs/~dynamic-names.yml

I've not contributed to any mustache projects before, so I've tried to keep my code change as concise and self-contained as possible, whilst getting all the relevant tests in the spec to pass.
(If you'd prefer something more verbose to aid code readability, I'm happy to adjust)


lucas42:
There's a problem with this change, which none of the tests have flagged.  Whilst it works the first time for rendering a given dynamic partial, the way the caching works is that'll it'll continue to render that same partial for subsequent calls, even if a different name is passed in.

lucas42:
Okay, I've managed to fix the caching problem.  Have added a test to verify that.
Though it does make this change more complicated than originally envisaged.  Would appreciate any feedback as I'm a bit out of my depth with the parsing logic.

PR #829, opened 2024-05-14
Upgrade modules used in github actions to latest supported versions

lucas42:
I noticed the github action workflows were displaying a lot of deprecation warnings.
Upgrading these modules should remove those warnings and ensure the CI jobs are running on supported version.


lucas42:

PR #830, opened 2024-05-14
Add newer versions of node to github actions

lucas42:
There's been quite a few new version of node released since the github actions config was written.
This adds all even versions (Long-term supported ones) to the github actions matrixes up to version 20.
It also uses version 20 for any jobs which run once.

Version 22 of node is also now available, but that seems to break the build, so will need a bit more investigation before adding.


lucas42:

PR #835, opened 2024-05-22
Upgrade action modules

lucas42:
Use a version of ngrok which runs on recent versions of Windows and MacOS. 
Switch to using the npm `overrides` property, to avoid needing to mantain a separate fork of `zuul-ngrok`


lucas42:
(Sorry, mixed up which branch I was on)

PR #836, opened 2024-05-22
Upgrade npm dev dependencies

lucas42:
Bump a bunch of dev dependency versions in package.json to more recent ones.

These aren't all the latest versions available, as some of those caused breaking changes in the tests.  Therefore, I've stopped at the most recent version I could find without a breaking change.

As part of this, I've had to move the node 10 tests into the "legacy" matrix.  (I'm not entirely sure which of the upgrades broke it specifically).

### DIFF
--- a/.github/workflows/verify.yml
+++ b/.github/workflows/verify.yml
@@ -20,7 +20,7 @@ jobs:
 
     strategy:
       matrix:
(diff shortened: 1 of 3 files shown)


lucas42:

Issue #807, opened 2022-10-21
Does the CLI alow for partials?


halloleo:
Does the library/CLI allow for mustache partials? If so, how? The help gives me:

    $ mustache -h
    Syntax: mustache <view> <template> [output]

And when I try something like 

    {{> file.html}} 

hoping `file.html` gets inlined, nothing is included.

phillipj:
Yepp, the README has an example of CLI & partials usage with the `-p` argument: https://github.com/janl/mustache.js#command-line-tool

PR #849, opened 2026-07-23
docs(cli): mention -p partials in usage syntax

Hashim1999164:
## Summary
- Updates the CLI usage message so it mentions the existing `-p` flag for partial templates.
- Addresses #807, where running `mustache -h` / missing args only showed `<view> <template> [output]` and made it look like partials were unsupported.

The README already documents `-p`; this just brings the runtime syntax hint in line with that.

## Test plan
- [x] Run `node bin/mustache` with no args and confirm stderr includes `-p`
- [x] Confirm existing `-p` partial rendering still works as before


lbesecker195 (approved):
Verified: `bin/mustache` already parses repeated `-p` arguments (lines 8 to 14), so the usage text was simply incomplete. LGTM.

